PR #10, opened 2016-07-26
Support nested submodules

mathias-baumann-sociomantic:
Extends the logic for the `make test` target to use the most appropriate `UnitTestRunner.d` module. 
Will use the most shallow path, e.g. if you have (number at the tail is the path deepness)

```
./submodules/ocean/src/ocean/core/UnitTestRunner.d 6
./submodules/lib1/submodules/ocean/src/ocean/core/UnitTestRunner.d 8
./submodules/lib2/submodules/ocean/src/ocean/core/UnitTestRunner.d 8
./submodules/lib3/submodules/ocean/src/tango/core/UnitTestRunner.d 8
```

it will take the first one in that list.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -463,15 +463,24 @@ $O/allunittests.d: $(UNITTEST_FILES)
 # 1. find any module named UnitTestRunner
 # 2. get the relative file path from `src` folder
 # 3. replace slashes with dots to get qualified module name
-TEST_RUNNER_MODULE=$(shell \
+TEST_RUNNER_MODULE?=$(shell \
 	  find ./ -path */src/*/core/UnitTestRunner.d \
 	| sed -n 's|^.\+/src/\(.\+/UnitTestRunner\).d$$|\1|p' \
 	| tr / .)
 
+# if no UnitTestRunner module is found and it was not overriden
+# from command line or Config.mak, use default D test runner
+# as a fallback
+ifeq ($(TEST_RUNNER_MODULE),)
+TEST_RUNNER_STRING=void main() {}
+else
+TEST_RUNNER_STRING=import $(TEST_RUNNER_MODULE);
+endif
+
 # General rule to build the unittest program using the UnitTestRunner
 $O/%unittests.d: $G/build-d-flags
 	$(call exec,printf 'module $(patsubst $O/%.d,%,$@);\n\
-		import $(TEST_RUNNER_MODULE); \
+		$(TEST_RUNNER_STRING)\n\
 		\n$(foreach f,$(filter %.d,$^),\
 		import $(subst /,.,$(patsubst $C/$(SRC)/%.d,%,$f));\n)' > \
 			$@,,gen)
--- a/Makd.mak
+++ b/Makd.mak
@@ -95,7 +95,7 @@ COLOR_OUT ?= $(COLOR_ERR)
 # To compile the D2 version, you can use make DVER=2
 # FIXME_IN_D2: This is only present as a transitional solution, it should be
 # removed after the D2 migration is done
-DVER := 1
+DVER ?= 1
 export MAKD_DVER := $(DVER)
 
 # Default D compiler (tries first with dmd1 and uses dmd if not present)
--- a/Makd.mak
+++ b/Makd.mak
@@ -461,13 +461,14 @@ UNITTEST_FILES += $(call find_files,.d,,$C/$(SRC),$(TEST_FILTER_OUT))
 $O/fastunittests.d: $(filter-out %_slowtest.d,$(UNITTEST_FILES))
 $O/allunittests.d: $(UNITTEST_FILES)
 
-# 1. find any module named UnitTestRunner
-# 2. get the relative file path from `src` folder
-# 3. replace slashes with dots to get qualified module name
-TEST_RUNNER_MODULE?=$(shell \
-	  find ./ -path */src/*/core/UnitTestRunner.d \
-	| sed -n 's|^.\+/src/\(.\+/UnitTestRunner\).d$$|\1|p' \
-	| tr / .)
+# Test runner module. Uses ocean if available
+TEST_RUNNER_MODULE=$(shell \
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
I think this might be getting too smart, this was done only to support the move of the `UnitTestRunner` from Tango to Ocean. Now that the variable `TEST_RUNNER_MODULE` is there, you can just override the module to use explicitly, and I think that's the best choice instead of guessing what the user wants to do.

I'm not sure what would be the most reasonable default though. Probably search in the submodule list for ocean and use the ocean one if it's there, otherwise use an empty `main()` so the default druntime unittest runner is used.

mathias-baumann-sociomantic:
My main motivation was that right now it simply breaks if there are several..

mathias-baumann-sociomantic:
Do you want me to change it to default to ocean and else an empty main?

mathias-baumann-sociomantic:
I mean, that might also be a bit tricky, what if someone forgot to clone the ocean submodule and does `make test` and it silently succeeds without actually testing anything.

Maybe the empty main shouldn't be actually empty but output something like "No tests were run" or so

leandro-lucarella-sociomantic:
> My main motivation was that right now it simply breaks if there are several..

Yes, I understand this, is just that the "discovery" idea was just a temporary hack in the first place, so this is why I think we shouldn't complicate it even more, but make it simpler again.

> Do you want me to change it to default to ocean and else an empty main?

Yes, I like that approach much better.

> I mean, that might also be a bit tricky, what if someone forgot to clone the ocean submodule and does make test and it silently succeeds without actually testing anything.

No, using an empty `main()` will still test everything, is just that the unit test runner used will be the default from druntime, which lacks many options of the ocean's one. This is already implemented and used by d1to2fix. So all that's needed is basically something like:

``` sh
ocean=`echo $(SUBMODULES) | grep -ox '.*\<ocean\>.*'`
if test -n "$ocean" &&
    TEST_RUNNER_MODULE=$ocean/src/ocean/core/UnitTestRunner.d
```

(untested)

mathias-baumann-sociomantic:
> No, using an empty main() will still test everything, is just that the unit test runner used will be the default from druntime

even the d1 one?

leandro-lucarella-sociomantic:
Yes.

mathias-baumann-sociomantic:
updated

Makd.mak
```diff
@@ -460,13 +460,18 @@ UNITTEST_FILES += $(call find_files,.d,,$C/$(SRC),$(TEST_FILTER_OUT))
 $O/fastunittests.d: $(filter-out %_slowtest.d,$(UNITTEST_FILES))
 $O/allunittests.d: $(UNITTEST_FILES)
 
-# 1. find any module named UnitTestRunner
-# 2. get the relative file path from `src` folder
-# 3. replace slashes with dots to get qualified module name
+
+# Default ocean test runner path
+OCEAN_TEST_RUNNER_MODULE=submodules/ocean/src/ocean/core/UnitTestRunner.d
```
leandro-lucarella-sociomantic: Mmm, it looks like you didn't have a look to my comment, this is not using git submodules to discover where/if ocean is used. I'm not saying that heuristic is flawless (it certainly isn't) but I think hardcoding the path like this is not a good idea, Makd doesn't impose any structure on where submodules should reside.

Makd.mak
```diff
@@ -460,13 +460,18 @@ UNITTEST_FILES += $(call find_files,.d,,$C/$(SRC),$(TEST_FILTER_OUT))
...
+# Test runner module. Uses ocean if available, else the default druntime runner
+# will be used
 TEST_RUNNER_MODULE=$(shell \
-	  find ./ -path */src/*/core/UnitTestRunner.d \
-	| sed -n 's|^.\+/src/\(.\+/UnitTestRunner\).d$$|\1|p' \
-	| tr / .)
+	if test -f "${OCEAN_TEST_RUNNER_MODULE}"; then \
+		echo ${OCEAN_TEST_RUNNER_MODULE}; \
+	else \
+		echo "object"; \
```
leandro-lucarella-sociomantic: Also, this should be empty if not found, if you check a bit further down, the `main()` is created only if the module is empty.

leandro-lucarella-sociomantic: Oh, I think you can't see it because is based on some old commit. Please rebase before continuing :)

mathias-baumann-sociomantic:
updooted

Makd.mak
```diff
@@ -461,13 +461,14 @@ UNITTEST_FILES += $(call find_files,.d,,$C/$(SRC),$(TEST_FILTER_OUT))
...
-# 2. get the relative file path from `src` folder
-# 3. replace slashes with dots to get qualified module name
-TEST_RUNNER_MODULE?=$(shell \
-	  find ./ -path */src/*/core/UnitTestRunner.d \
-	| sed -n 's|^.\+/src/\(.\+/UnitTestRunner\).d$$|\1|p' \
-	| tr / .)
+# Test runner module. Uses ocean if available
+TEST_RUNNER_MODULE=$(shell \
+	for sub in ${SUBMODULES}; do \
+		if test "$$(basename $$sub)" = "ocean"; then \
```
leandro-lucarella-sociomantic, line 467: mmm, this is a bit more restricted than what I proposed but I guess it will do.

leandro-lucarella-sociomantic:
Mmm, this is also based on v1.4.x but it seems like you rebased against v1.5.x. I think it would be best to reopen against v1.5.x, since you are using new features from it (I don't think it's worth porting back to v1.4.x)

mathias-baumann-sociomantic:
https://github.com/sociomantic-tsunami/makd/pull/13

Makd.mak
mathias-baumann-sociomantic, line 467: Maybe I just didn't understand you ;)

mathias-baumann-sociomantic, line 467: Oh, i didn't see the example code ^^

PR #18, opened 2016-09-19
Also consider unannotated tags in version

mathias-baumann-sociomantic:


leandro-lucarella-sociomantic:
This was suggested before and I don't like it because tags for releases should be annotated (so you have a git object associated with the tag, and you can, for example, cryptographically sign them). It also have a lot of sense to include release notes in the message for example. *Lightweight* tags are useful to mark stuff in the local repository, they are just like a bookmark, so if Makd interpret them as releases too, you lose that distinction, because it will use the latest tag it can find. Honestly I don't know why git support unannotated tags really, that's just a branch. Well, maybe just to mark an abandoned branch you don't want people to keep committing to.

And then, if `git describe` doesn't consider lightweight tags when describing the version of the repository is for a reason, and if we do, we'll confuse users that are used to how git works and expect lightweight tags to remain... lightweight.

All that said, I will give this some time for discussion anyway... But if a strong argument in favour of this comes up, I'll close as `wontfix`.

leandro-lucarella-sociomantic (changes requested):
This was suggested before and I don't like it because tags for releases should be annotated (so you have a git object associated with the tag, and you can, for example, cryptographically sign them). It also have a lot of sense to include release notes in the message for example. _Lightweight_ tags are useful to mark stuff in the local repository, they are just like a bookmark, so if Makd interpret them as releases too, you lose that distinction, because it will use the latest tag it can find. Honestly I don't know why git support unannotated tags really, that's just a branch. Well, maybe just to mark an abandoned branch you don't want people to keep committing to.

And then, if `git describe` doesn't consider lightweight tags when describing the version of the repository is for a reason, and if we do, we'll confuse users that are used to how git works and expect lightweight tags to remain... lightweight.

All that said, I will give this some time for discussion anyway... But if a strong argument in favour of this comes up, I'll close as `wontfix`.

leandro-lucarella-sociomantic:
Closing because of the lack of feedback.

PR #22, opened 2016-09-29
Remove deprecated Version alias to versionInfo

mathias-lang-sociomantic:
Quite a breaking change right ? :D

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,10 +1,2 @@
 New Features
 ============
-
-* `mkpkg`: `FUN.autodeps()` now accepts multiple arguments. The set of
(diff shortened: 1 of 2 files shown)


leandro-lucarella-sociomantic (changes requested):
Update release notes please. Since this is not a typical library where we can say "all deprecated features removed" I prefer to be explicit about this change in the release notes :)

mihails-strasuns-sociomantic:
Ping @mathias-lang-sociomantic

mathias-lang-sociomantic:
Done. Although I should probably not be the one clearing RELEASE_NOTES

mihails-strasuns-sociomantic:
LGTM, thanks!

PR #24, opened 2016-10-10
Export DC to mkversion.sh run in shell function

### LINKED ISSUE #23: mkversion.sh refers to 'dmd1'
```
/mnt/shared/projects/audio/wavegen/submodules/makd/mkversion.sh: line 12: dmd1: command not found
make: Nothing to be done for 'all'.
```

Which comes from here:
https://github.com/sociomantic-tsunami/makd/blob/497a15fe5d6fe8a5079054b21df6295ab5fd1b22/mkversion.sh#L12

I thought [DC should be exported by Makd](https://github.com/sociomantic-tsunami/makd/blob/497a15fe5d6fe8a5079054b21df6295ab5fd1b22/Makd.mak#L103-L109), but apparently it isn't.

Also, there is a small typo in the usage:

> this script doesn't **get** care of quoting, so if you use any 'special'

nemanja-boric-sociomantic:
Shell function doesn't inherit variables exported
in Makefile, so explicit passing of `DC` is performed.

Fixes #23

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -211,7 +211,11 @@ VERSION_FILE := $(GS)/Version.d
 # `-` in the dirty separator is also replaced with `~`. This is done for
 # improved Debian version comparison.
 # Example: v1.2.1-4-gea7105b-dirty -> 1.2.1+4~dirty.20160202160552~ea7105b
-PKGVERSION := $(shell $(MAKD_PATH)/mkversion.sh -p | \
+#
+# Shell function doesn't inherit variables exported in Makefile, so
+# explicit passing of `DC` is performed (https://bugs.debian.org/184864)
+#
+PKGVERSION := $(shell DC="$(DC)" $(MAKD_PATH)/mkversion.sh -p | \
                 sed 's/^v\(.*\)/\1/' | \
                 sed 's/^\(.*\)-\([0-9]\+\)-g\([0-9a-f]\+\)/\1+\2~\3/' | \
                 sed 's/^\([^0-9]\)/0\1/' | \
--- a/mkversion.sh
+++ b/mkversion.sh
@@ -9,7 +9,7 @@ rev_file=src/Version.d
 author="`git config user.name`"
 
 # Get compiler version
-compiler="`${DC:-dmd1} | head -1`"
+compiler="`${DC:-dmd} | head -1`"
 
 # Get the current date (might be overridden by command-line options later)
 date=$(date -u +"%Y-%m-%d %H:%M:%S %Z")


leandro-lucarella-sociomantic (changes requested):
`DC` is supposed to be exported, see https://github.com/sociomantic-tsunami/makd/blob/da527f1e33e00f1a0d02745d0fe366e602639c83/Makd.mak#L106-L108. My guess is `?=` is what is making it fail on some circumstances, I guess something like:

``` make
DC ?= ...
export DC
```

Should fix it.

nemanja-boric-sociomantic:
@leandro-lucarella-sociomantic I'm not sure that `DC` is exported to the subshell when called with `$(shell`

nemanja-boric-sociomantic:
https://bugs.debian.org/184864

nemanja-boric-sociomantic:
But, I'll try with `export DC`

nemanja-boric-sociomantic:
Yeah, it doesn't work.

leandro-lucarella-sociomantic:
:cry:

leandro-lucarella-sociomantic:
Add a comment with a link to the issue maybe?

leandro-lucarella-sociomantic (approved):

nemanja-boric-sociomantic:
Updated with the issue link in the commit message.

mathias-lang-sociomantic:
I'd also make the default in `mkversion.sh` refers to `dmd` instead of DMD1.
It shouldn't affect us, but will make the script more usable for upstream.

leandro-lucarella-sociomantic (changes requested):
> I'd also make the default in mkversion.sh refers to dmd instead of DMD1.
> It shouldn't affect us, but will make the script more usable for upstream.

Yeah, I haven't saw the original issue before, it definitely make sense to change the default too (as a separate commit because they are 2 different things, but I think both are needed to fix the original issue properly).

nemanja-boric-sociomantic:
Updated.

mathias-lang-sociomantic (approved):
LGTM, thanks :)

leandro-lucarella-sociomantic:
> Updated with the issue link in the commit message.

Mmm, I meant a comment in the code (too), as I have the feeling someone (like me) would see explicit env var passing in the future and think is redundant and remove it :)

nemanja-boric-sociomantic:
You're such a pain :-).

Updated.

leandro-lucarella-sociomantic (approved):

leandro-lucarella-sociomantic:
> You're such a pain :-).

:kissing_heart:

Issue #11, opened 2016-08-01
Modify d1to2fix support after new version is released


mihails-strasuns-sociomantic:
- collect all modules to be converted into list file (https://github.com/sociomantic-tsunami/d1to2fix/pull/4)
- modify automatic conversion target so that all submodules are converted in one step with main sources (need for performance reasons because of https://github.com/sociomantic-tsunami/d1to2fix/pull/3)

PR #25, opened 2016-10-12
Update Makd to use d1to2fix v0.7.0 features

mihails-strasuns-sociomantic:
Fixes https://github.com/sociomantic-tsunami/makd/issues/11

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -386,8 +386,8 @@ endef
 RDMDFLAGS ?= --force --compiler=$(DC)
 
 # Default dmd flags
-override DFLAGS += -I$(GS) -I./$(SRC) $(foreach dep,$(SUBMODULES), -I./$(dep)/$(SRC))
-
+DIMPORTPATHS := -I$(GS) -I./$(SRC) $(foreach dep,$(SUBMODULES), -I./$(dep)/$(SRC))
+override DFLAGS += $(DIMPORTPATHS)
 
 # Include the user's makefile, Build.mak
 #########################################
@@ -645,7 +645,7 @@ d2conv: $O/d2conv.stamp
 $O/d2conv.stamp: $C
 	$Vfind $C -type f -regex '^.+\.d$$' > $@
 ifeq "$(shell d1to2fix --help | grep -- --input)" ""
-	$(call exec, d1to2fix --fatal --input=$@)
+	$(call exec, d1to2fix $(DIMPORTPATHS) --fatal --input=$@)
 else
 	$(call exec, d1to2fix --fatal `cat $@`)
 endif
--- a/Makd.mak
+++ b/Makd.mak
@@ -640,19 +640,15 @@ fasttest: $(fasttest)
 
 
 .PHONY: d2conv
-d2conv: $O/d2conv.stamp \
-	$(foreach s,$(SUBMODULES),$O/d2conv/$s.stamp)
+d2conv: $O/d2conv.stamp
 
 $O/d2conv.stamp: $C
-	$(call exec,git ls-files | grep '\.d$$' | \
-		xargs --no-run-if-empty d1to2fix,.,d1to2fix)
-
-$O/d2conv/%.stamp: $C/%
-	$(call exec,cd $<; git ls-files | grep '\.d$$' | \
-		xargs --no-run-if-empty d1to2fix,$<,d1to2fix)
-	$Vmkdir -p $(dir $@)
-	$Vtouch $@
-
+	$Vfind $C -type f -regex '^.+\.d$$' > $@
+ifeq "$(shell d1to2fix --help | grep -- --input)" ""
+	$(call exec, d1to2fix --fatal --input=$@)
+else
+	$(call exec, d1to2fix --fatal `cat $@`)
+endif
 
 # Automatic dependency handling
 ################################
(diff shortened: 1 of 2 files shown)


mihails-strasuns-sociomantic:
As this won't work without recent enough d1to2fix, it may be better to put it to v2.x.x though

leandro-lucarella-sociomantic:
> As this won't work without recent enough d1to2fix, it may be better to put it to v2.x.x though

Mmmm, what about making it optional? Can't we detect d1to2fix version and act based on that?

mihails-strasuns-sociomantic:
Possible, but would require duplicating all relevant `.stamp` / `.list` rules with a different definitions - rather fragile and messy. I'd prefer a clean split if possible.

Makd.mak
```diff
@@ -640,19 +640,19 @@ fasttest: $(fasttest)
...
-$O/d2conv.stamp: $C
-	$(call exec,git ls-files | grep '\.d$$' | \
-		xargs --no-run-if-empty d1to2fix,.,d1to2fix)
-
-$O/d2conv/%.stamp: $C/%
-	$(call exec,cd $<; git ls-files | grep '\.d$$' | \
-		xargs --no-run-if-empty d1to2fix,$<,d1to2fix)
-	$Vmkdir -p $(dir $@)
+$O/d2conv.stamp: $O/d2conv.list
+	$(call exec, d1to2fix --fatal --input=$^, $^)
```
leandro-lucarella-sociomantic: You should use `$<` instead, you only want the first pre-requisite (just in case you add more in the future)

Makd.mak
```diff
@@ -640,19 +640,19 @@ fasttest: $(fasttest)
...
 	$Vtouch $@
 
+$O/d2conv.list: $(foreach s,$(SUBMODULES), $O/d2conv/$s.list)
+	$Vgit ls-files | { grep '\.d$$' || true; } > $@
+	$Vcat $^ >> $@
```
leandro-lucarella-sociomantic: I would use `find_files` here for consistency. Also you could combine the stamp with the list, as the stamp is just an artefact created because the building doesn't produce any real output, but in this case the list file can be used as output.

On the other hand, I think it would make more sense for d1to2fix to support `--stdin` to just get the files via stdin, there is no much of a point on using a file really :D

leandro-lucarella-sociomantic:
> Possible, but would require duplicating all relevant .stamp / .list rules with a different definitions - rather fragile and messy. I'd prefer a clean split if possible.

Messy, maybe, fragile, I don't see how. I think I might be OK with a clean cut, but I would like Makd to error if d1to2fix is old with a nice message (not the error message of d1to2fix complaining about unknown flags).

But if this is somehow urgent, it will get very low acceptance, so nobody will update to v2 and it will be like this change wasn't there. This is why a non-breaking update path seems to be more profitable for me.

mihails-strasuns-sociomantic:
After some thinking, I have realized that with "convert all at once" approach actual makefile target can be much much more simple and preserve backwards compatibility at the same time.

Updated and should be OK for v1.x.x now too

Makd.mak
```diff
@@ -640,19 +640,15 @@ fasttest: $(fasttest)
...
-		xargs --no-run-if-empty d1to2fix,.,d1to2fix)
-
-$O/d2conv/%.stamp: $C/%
-	$(call exec,cd $<; git ls-files | grep '\.d$$' | \
-		xargs --no-run-if-empty d1to2fix,$<,d1to2fix)
-	$Vmkdir -p $(dir $@)
-	$Vtouch $@
-
+	$Vfind $C -type f -regex '^.+\.d$$' > $@
+ifeq "$(shell d1to2fix --help | grep '\-\-input')" ""
```
nemanja-boric-sociomantic: :-)

Makd.mak
```diff
@@ -386,8 +386,8 @@ endef
 RDMDFLAGS ?= --force --compiler=$(DC)
```
nemanja-boric-sociomantic, line 386: I think the commit message is not right here?

nemanja-boric-sociomantic: I mean - what's the connection between new d1to2fix feature and imports?

Makd.mak
nemanja-boric-sociomantic: I'm not if you would want to use `grep -q` here and to test the return value of `grep`.

nemanja-boric-sociomantic: (seems more idiomatic)

mihails-strasuns-sociomantic: Yes, but `ifeq` can only compare strings afaik so it felt like overcomplication

nemanja-boric-sociomantic: fair enought

Makd.mak
mihails-strasuns-sociomantic: New d1to2fix uses limited semantic analysis provided by dsymbol library. For that to work and resolve symbols properly, it needs know all import paths same as compiler does. If import paths are not available, it won't be able to detect any aliased delegates.

nemanja-boric-sociomantic: Cool :+1: Do you care to put that in the commit message maybe (to require imports to do this).

Makd.mak
llucax: Also `grep -- --input`.

mihails-strasuns-sociomantic:
Enhanced commit wording and used @leandro-lucarella-sociomantic suggestion

PR #26, opened 2016-10-13
Minimal support creating non-deb packages

mihails-strasuns-sociomantic:
At this step changes are just enough to be able to create
Arch Linux pacman packages.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -128,6 +128,9 @@ FPM ?= fpm
 # Default dot binary location
 DOT ?= dot
 
+# Default package type
+PKGTYPE ?= deb
+
 # Default flags to pass to graph-deps (see --help for details)
 GRAPH_DEPS_FLAGS ?= -c -C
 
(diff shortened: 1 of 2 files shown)


Makd.mak
```diff
@@ -128,17 +128,21 @@ FPM ?= fpm
...
 # (defined lazily so we can use target variables)
 ifndef PKG_DEFAULTS
-PKG_DEFAULTS = -D-f -D-sdir -D-tdeb -D--deb-changelog="$O/changelog.Debian" \
+DISTRO_CODENAME="$(shell lsb_release -rs | grep rolling || lsb_release -cs)"
+PKG_DEFAULTS = -D-f -D-sdir -D-t$(PKGTYPE) -D--deb-changelog="$O/changelog.Debian" \
 		-D--version="$(PKGVERSION)" \
-		-D--iteration="$(shell lsb_release -cs)" \
-		-d lsb_release="$(shell lsb_release -cs)"
+		-D--iteration="$(DISTRO_CODENAME)" \
+		-d lsb_release="$(DISTRO_CODENAME)"
```
leandro-lucarella-sociomantic, line 145: What about having some sort of _switch_ based on `lsb_release -is`, if is `Ubuntu`/`Debian` then do this, if is _arch_, do that, else error?

leandro-lucarella-sociomantic (commented):
What about autodeps? It will fail in another environment, right?

Makd.mak
mihails-strasuns-sociomantic: I think current version is actually more reliable. It will work with any distro that reports itself as rolling release in `lsb_release -rs`, not just Arch.

mihails-strasuns-sociomantic:
> What about autodeps? It will fail in another environment, right?

Yep, this is just the minimal amount of changes needed to generate package.

leandro-lucarella-sociomantic:
I feel `mkpkg` should at least fail gracefully with a nice "unsupported" error message.

leandro-lucarella-sociomantic (changes requested):
I feel like having a proper error message en `mkpkg` saying that `autodeps()` is not supported if `type=arch` is a prerequisite to merge this. We can have the MakD crashing with an ugly backtrace with cryptic errors if somebody tries to use `arch` + `autodeps()`.

mihails-strasuns-sociomantic:
Something like e905bf4?

leandro-lucarella-sociomantic (approved):

PR #29, opened 2016-11-11
Fix wrong if branches of d1to2fix feature check

mihails-strasuns-sociomantic:
Actions upon each branch were exchanged to what they must have
been. Resulted in misbehaving D2 conversion not using new d1to2fix features.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -652,9 +652,9 @@ d2conv: $O/d2conv.stamp
 $O/d2conv.stamp: $C
 	$Vfind $C -type f -regex '^.+\.d$$' > $@
 ifeq "$(shell d1to2fix --help | grep -- --input)" ""
-	$(call exec, d1to2fix $(DIMPORTPATHS) --fatal --input=$@)
-else
 	$(call exec, d1to2fix --fatal `cat $@`)
+else
+	$(call exec, d1to2fix $(DIMPORTPATHS) --fatal --input=$@)
 endif
 
 # Automatic dependency handling
--- a/Makd.mak
+++ b/Makd.mak
@@ -651,7 +651,7 @@ d2conv: $O/d2conv.stamp
 
 $O/d2conv.stamp: $C
 	$Vfind $C -type f -regex '^.+\.d$$' > $@
-ifeq "$(shell d1to2fix --help | grep -- --input)" ""
+ifeq "$(shell d1to2fix --help 2>/dev/null | grep -- --input)" ""
 	$(call exec, d1to2fix --fatal `cat $@`)
 else
 	$(call exec, d1to2fix $(DIMPORTPATHS) --fatal --input=$@)


Makd.mak
```diff
@@ -652,9 +652,9 @@ d2conv: $O/d2conv.stamp
 $O/d2conv.stamp: $C
 	$Vfind $C -type f -regex '^.+\.d$$' > $@
 ifeq "$(shell d1to2fix --help | grep -- --input)" ""
```
leandro-lucarella-sociomantic: Unrelated, but this also makes make to fail if no `d1to2fix` is present, even for stuff that never uses it. Maybe it would be nice to fix in this release too.

The check could actually be moved to the shell instead of doing it via Make:

``` make
$O/d2conv.stamp: $C
    $Vfind $C -type f -regex '^.+\.d$$' > $@
    $(call exec, d1to2fix --help | grep -q -- --input &&
            d1to2fix $(DIMPORTPATHS) --fatal --input=$@ ||
            d1to2fix --fatal `cat $@`)
```

This way `d1to2fix` will only be used when needed.

mathias-lang-sociomantic (approved):
Dammit. LGTM.

leandro-lucarella-sociomantic (commented):
Should we fix both in one go?

gavin-norman-sociomantic:
Works in my test with the dhtnode :+1:

Makd.mak
mihails-strasuns-sociomantic: I like how existing version looks nice with `make d2conv V=1`, how about this instead?

```
$O/d2conv.stamp: $C
    $Vfind $C -type f -regex '^.+\.d$$' > $@
ifneq "$(shell d1to2fix --help | grep d1to2fix [OPTIONS] [FILES])" ""
ifeq "$(shell d1to2fix --help | grep -- --input)" ""
    $(call exec, d1to2fix --fatal `cat $@`)
else
    $(call exec, d1to2fix $(DIMPORTPATHS) --fatal --input=$@)
endif
endif
```

mihails-strasuns-sociomantic: Or probably I can simply use bash if and output suppression, gonna try it..

leandro-lucarella-sociomantic: If you want to keep the current structure, the easiest fix is: `$(shell d12tofix --help 2>/dev/null | grep -- --input)`. I think is a bit ugly to try to run `d1to2fix` just by parsing the `Makefile`, but I won't insist a lot on the which solution to use. Is a transitional step anyway...

mihails-strasuns-sociomantic:
Updated with cca296d

Makd.mak
```diff
@@ -651,11 +651,10 @@ d2conv: $O/d2conv.stamp
 
 $O/d2conv.stamp: $C
 	$Vfind $C -type f -regex '^.+\.d$$' > $@
-ifeq "$(shell d1to2fix --help | grep -- --input)" ""
+	@if d1to2fix --help | grep -- --input; then \
```
leandro-lucarella-sociomantic: I guess you explicitly want to silence those `if` in verbose more, right? Otherwise you should use `$V`.

mihails-strasuns-sociomantic: Yes. I want only the statement matching current d1to1fix version to be visible in verbose mode

Makd.mak
```diff
@@ -651,11 +651,10 @@ d2conv: $O/d2conv.stamp
...
+	$(call exec, d1to2fix $(DIMPORTPATHS) --fatal --input=$@) \
+	@else \
```
leandro-lucarella-sociomantic: Now, wait, this wont work. That's not how make works.

Each line will invoke a shell (of course lines with _line continuation_ `\` are considered only one line), and that shell command can optionally start with `@` to avoid printing the command. So this `@else` will be passed to the shell literally as `@else`.

Or did you actually tried it and worked as you expected?

mihails-strasuns-sociomantic: > Or did you actually tried it and worked as you expected?

It did :D

mihails-strasuns-sociomantic:
```
$ make d2conv V=2
find /home/dicebot/sociomantic/ocean -type f -regex '^.+\.d$' > /home/dicebot/sociomantic/ocean/build/devel/tmp/d2conv.stamp
d1to2fix -I/home/dicebot/sociomantic/ocean/build/devel/include -I./src  -I./submodules/makd/src --fatal --input=/home/dicebot/sociomantic/ocean/build/devel/tmp/d2conv.stamp 
```

leandro-lucarella-sociomantic: Something is very fishy about this, because you should also see the output of the `d1to2fix --help | grep -- --input` command. I'll test.

leandro-lucarella-sociomantic: Without `V=1` I get `/bin/sh: 4: Syntax error: end of file unexpected (expecting "fi")`. This is definitely broken, the statement before `else` also must end with `;` when its in one line. I don't know because of what black magic is working for you, but this is definitely broken.

leandro-lucarella-sociomantic: Just leave it as it was before and add the `2>/dev/null` to suppress errors if you want your clean verbose (but not **too** verbose) output :D

mihails-strasuns-sociomantic: Done

leandro-lucarella-sociomantic:
This pull request has been rebased via `git hub pull rebase`. Original pull request HEAD was e271479cb1d1ab4285ebc5f7b9a2ed53f82bde2c, new (rebased) HEAD is 91dc6204b83298a6c431bfc993524f00877a937d

PR #30, opened 2016-11-22
`make pkg` cannot handle UTF8 in commit messages

leandro-lucarella-sociomantic:
It seems that if commit messages include UTF8 characters, `make pgk` will fail with an error message along the lines of the following:
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 96: ordinal not in range(128)
```

This seems to come down to the following command in the `git-deblog` script:
https://github.com/sociomantic-tsunami/makd/blob/1446612df547465d402d349744e8cab0f1358dbf/git-deblog#L125

### DIFF
--- a/git-deblog
+++ b/git-deblog
@@ -169,7 +169,7 @@ def parse_args():
             "commit is used if present)")
     parser.add_argument('-d', '--dist-name', default=get_dist(),
             help="specifies the distribution name (by default obtained from "
-            "lsb_release")
+            "lsb_release)")
     parser.add_argument('-u', '--urgency', default='medium',
             choices='low medium high emergency critical'.split(),
             help="specifies the distribution name (by default obtained from "
--- a/git-deblog
+++ b/git-deblog
@@ -5,18 +5,28 @@
 # (See accompanying file LICENSE or copy at http://www.boost.org/LICENSE_1_0.txt)
 
 import sys
-import locale
 from datetime import datetime
 from subprocess import CalledProcessError, PIPE
-
-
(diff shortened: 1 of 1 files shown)


leandro-lucarella-sociomantic:
Fix attached.

leandro-lucarella-sociomantic:
Rebased. Please give this a bit of attention since we need the fix soon.

git-deblog
```diff
@@ -49,8 +59,11 @@ def main():
 
 
 def run(*args, **kwargs):
+    from locale import getpreferredencoding
     from subprocess import check_output
-    return check_output(*args, **kwargs).strip().decode(ENC)
+    encoding = getpreferredencoding()
```
nemanja-boric-sociomantic, line 64: Is this then for everything else except `git log`?

leandro-lucarella-sociomantic: Yeah, is a tricky issue always with encodings... I'm trusting Python and the environment to set the correct encodings, but I have the feeling we'll have this problem also with `git config` stuff (an author name in UTF-8 and a poorly configured environment). I'll add options to override this via command-line option in the (very near) future.

nemanja-boric-sociomantic:
Enjoy.

joseph-wakeling-sociomantic:
Thanks @leandro-lucarella-sociomantic @nemanja-boric-sociomantic ! :-)

joseph-wakeling-sociomantic:
Looks like the issue persists, but this time in `get_git_author` :-(

PR #46, opened 2017-02-10
mkpkg: Fix _expand_list (autodeps and mapbins)

leandro-lucarella-sociomantic:
The _expand_list was misbehaving when passing a list to autodeps and
mapbins instead of using the old way to pass multiple arguments.

At least this is not a regression, it just make the new feature
unusable.

We **really** need to start thinking about adding tests to this :(

### DIFF
--- a/mkpkg
+++ b/mkpkg
@@ -124,7 +124,7 @@ class Functions:
         s = set()
         for f in files:
             if type(f) in (list, tuple, set, frozenset):
-                s.update(path + f)
+                s.update([path + ff for ff in f])
             else:
                 s.add(path + f)
         return s


mkpkg
```diff
@@ -124,7 +124,7 @@ class Functions:
         s = set()
         for f in files:
             if type(f) in (list, tuple, set, frozenset):
-                s.update(path + f)
+                s.update([path + ff for ff in f])
```
nemanja-boric-sociomantic, line 127: I see.

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
> We really need to start thinking about adding tests to this :(

+100. It's really hard to spot errors like this while reviewing, especially for people that don't write python every day.

leandro-lucarella-sociomantic:
> +100. It's really hard to spot errors like this while reviewing, especially for people that don't write python every day.

Yeah, it really sucks, is like coding blind-folded :(

PR #52, opened 2017-02-27
Fix broken link to makeit

andreas-bok-sociomantic:


llucax (approved):
The old link works too but better to update sooner than later.

PR #53, opened 2017-02-28
pkg: Force rebuild if PKG_SUFFIX changed

leandro-lucarella-sociomantic:
Usually when the `PKG_SUFFIX` changed, the package building should be forced. This didn't happen so if the user did a `make pkg` and then `make pkg PKG_SUFFIX=-test` for example, `make` would say that everything is up to date when it really isn't.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -447,7 +447,7 @@ pkg: $(patsubst $(PKG)/%.pkg,$O/pkg-%.stamp,$(wildcard $(PKG)/*.pkg))
 # Target to build a package based on the fpm definition (old packages are
 # removed to avoid infinite pollution of the build directory, as every package
 # is a different file)
-$O/pkg-%.stamp: $(PKG)/%.pkg
+$O/pkg-%.stamp: $(PKG)/%.pkg $G/pkg-flags
 	$(PKG_PREBUILD)
 	$(call exec,PYTHONDONTWRITEBYTECODE=1 $(MAKD_PATH)/mkpkg \
 		$(if $V,,-vv) -D-p"$P" -F "$(FPM)" \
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic (approved):

leandro-lucarella-sociomantic:
I did some testing, but maybe other people can test too to avoid another burst of point releases fixing regressions? :P

/cc @joseph-wakeling-sociomantic

leandro-lucarella-sociomantic:
ping @nemanja-boric-sociomantic @joseph-wakeling-sociomantic testing? :)

nemanja-boric-sociomantic:
I'll test today.

nemanja-boric-sociomantic:
For me it looks like this always force rebuilds.

nemanja-boric-sociomantic:
(even if `PKG_SUFFIX` has not changed)

leandro-lucarella-sociomantic:
That's weird. It seems to work fork me.

nemanja-boric-sociomantic:
You're happy to come and see what's going on.

leandro-lucarella-sociomantic:
OK, the problem seems to be `PKGVERSION` when the repo is dirty, because the version will get a timestamp that changes all the time. Since `PKGVERSION` is not supposed to be defined by the user (at least not in the common case), I think we can omit it from the flags to monitor.

leandro-lucarella-sociomantic:
Updated.

nemanja-boric-sociomantic:
Still doesn't work. There's also a flag that's passed to dmd:

```
 ~ -D-f -D-sdir -D-tdeb -D--deb-changelog="/home/nemanjaboric/work/xxx/build/devel/tmp/changelog.Debian" -D--version="1.8.3+35~dirty.20170317144224~b0188225" -D--iteration=""xenial"" -d lsb_release=""xenial"" ~ @rm -f /home/nemanjaboric/work/xxx/build/devel/pkg/_*.deb ; /home/nemanjaboric/work/xxx/submodules/makd/git-deblog -i "1.8.3+35~dirty.20170317144224~b0188225" "" --since '1 year ago' > /home/nemanjaboric/work/xxx/build/devel/tmp/changelog.Debian ~ fpm ~ /home/nemanjaboric/work/xxx/build/devel/bin ~ /home/nemanjaboric/work/xxx/build/devel ~
```

```
-D--version="1.8.3+35~dirty.20170317144224~b0188225" 
```

includes timestamp

leandro-lucarella-sociomantic:
To `fpm` you mean I guess... :(

nemanja-boric-sociomantic:
Correct, sorry.

leandro-lucarella-sociomantic:
OK, that appears in `PKG_DEFAULTS` that is also thought to be only defined in `Config.mak` if you want to do things radically differently (which was actually never tested), so I think it's safe to remove that too.

I will cut it to whatever is necessary and safe. The most important thing is to rebuild when `PKG_SUFFIX` changes, because that's what the user will likely change *often*. I left `FPM` too, in case some user might like to try building with different versions or something (and because it should be safe too).

leandro-lucarella-sociomantic:
(updated again)

nemanja-boric-sociomantic:
looks good and works good :+1:

PR #55, opened 2017-03-01
Remove unused D_GC variable and documentation

mathias-lang-sociomantic:
This doesn't seem to be used anywhere, hence why I targeted v1.x.x. Let me know if you want to re-target.

OT: v1.x.x should probably be merged into v2.x.x (which is 50 commits behind).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,3 +3,6 @@ Migration Instructions
 
 * The `Version` alias to `versionInfo` provided by the `Version` module has been removed.
   All remaining usage can be replaced with `versionInfo`.
+
+* The `D_GC` variable was removed as it wasn't used internally anymore.
+  If you made use of it anywhere, you should define it yourself (in `Config.mak` for example) from now on.
--- a/Makd.mak
+++ b/Makd.mak
@@ -111,11 +111,6 @@ endif
 # Default rdmd binary to use (same as with dmd)
 RDMD ?= rdmd
 
-# Garbage Collector to use
-# (exported because other programs might use the variable)
-D_GC ?= cdgc
-export D_GC
-
 # Default documentation generator tool to use (harbored-mod)
 HMOD ?= hmod
 
--- a/README.rst
+++ b/README.rst
@@ -361,7 +361,6 @@ Variables you might want to override
 * Program location variables: ``DC`` is the D compiler to use, you can build
   your project with a different DMD by using ``make
   DC=/usr/bin/experimental-dmd`` for example. Same for ``RDMD`` and ``FPM``.
-* ``D_GC`` to change the default (cdgc) GC implementation to use.
 * Less likely you might want to override the ``DFLAGS``, ``RDMDFLAGS`` or
   ``FPMFLAGS``, but usually there are better methods to do that instead.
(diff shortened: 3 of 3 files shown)


nemanja-boric-sociomantic:
Why not rename it to `DRT_GCOPT`, as it is done in druntime?

nemanja-boric-sociomantic:
Ah, that was removed as well.

leandro-lucarella-sociomantic:
It is used (in D1), but in a very stealth way: https://github.com/sociomantic-tsunami/tangort/blob/v1.x.x/dmd1.conf#L4

That said, since the config file shipped in tangort now also provides a default (this was probably added before `dmd1.conf` supported adding a default value), I guess it's OK to remove the default. The problem is without the `export` line in MakD, you can propagate the environment variable to the child process (`dmd1`) and thus you can't override it via `make`.

We can still use `D_GC=basic make` I think, and that would be propagated to `dmd1`, so we have a workaround (please test :) ), but this would qualify as a breaking change then, so I would favour going to v2.x.x branch (which I will get up to date soon).

leandro-lucarella-sociomantic:
> Why not rename it to DRT_GCOPT, as it is done in druntime?

They are not the same, `D_GC` is just a hack in the `dmd1.conf` shipped with tangort.

mathias-lang-sociomantic:
Didn't we remove the basic GC from tangort long ago though ?

leandro-lucarella-sociomantic:
> Didn't we remove the basic GC from tangort long ago though ?

Oh, [it looks like we did](https://github.com/sociomantic-tsunami/tangort/tree/v1.x.x/src/gc)! I do not recall at all that happened. Then we need to clean the `dmd1.conf` shipped by tangort too :grimacing:

leandro-lucarella-sociomantic:
Still, technically there are visible changes to the user, and since this change is not really buying us nothing, I would target v2.x.x anyway. Better be safe than sorry.

leandro-lucarella-sociomantic:
This also means adding a "Migration Instructions" note in the RELEASE NOTES.

leandro-lucarella-sociomantic:
#56

leandro-lucarella-sociomantic (changes requested):
Needs release notes and blocked by #56.

leandro-lucarella-sociomantic:
Unblocked, #56 merged.

mathias-lang-sociomantic:
Updated

RELEASE_NOTES.md
```diff
@@ -3,3 +3,5 @@ Migration Instructions
 
 * The `Version` alias to `versionInfo` provided by the `Version` module has been removed.
   All remaining usage can be replaced with `versionInfo`.
+
+* The `D_GC` variable available to `Build.mak` has been removed as there is only one GC available.
```
leandro-lucarella-sociomantic: I would rephrase this with something more in the lines of a migration instruction, like:

> * The `D_GC` variable was removed as it wasn't used any more internally. if you made use of it anywhere, you should define it yourself (in `Config.mak` for example) from now on.

leandro-lucarella-sociomantic (changes requested):

mathias-lang-sociomantic:
Updated

leandro-lucarella-sociomantic (approved):

PR #57, opened 2017-03-10
Add a new mapfiles function

gkotian:
The current `mapbins()` function is a convenient way to map local files into the package, but there is no reason for it to be limited to only mapping binaries. For this reason, a new `mapfiles()` function is introduced.

However, when mapping non-binaries, `VAR.suffix` may not be needed to be appended. So we need to also provide a way to control this.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -24,3 +24,5 @@ Deprecations
 ============
 
 * `VAR.name` is deprecated, please use `VAR.fullname` instead.
+
+* `FUN.mapbins()` is deprecated, please use `FUN.mapfiles()` instead.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,6 +14,11 @@ New Features
   description), `prolog` (inserted before the long description) and an `epilog`
   (appended at the end of the long description.
 
+* `FUN.mapfiles()`
+
+  A simple function to ease specifying files to include in the package (with the
+  ability to control whether `VAR.suffix` is appended to each destination file
+  using the named argument `append_suffix`).
 
 Deprecations
 ============
--- a/mkpkg
+++ b/mkpkg
@@ -166,6 +166,7 @@ class Functions:
                     for b in self._expand_list(bins)])
 
     def mapbins(self, src, dst, *bins):
+        warnf("'FUN.mapbins()' is deprecated, please use 'FUN.mapfiles()' instead")
         return sorted(['{src}/{bin}={dst}/{bin}{suffix}'.format(
                 src=src, dst=dst, bin=b, suffix=self.pkg.vars['suffix'])
                     for b in self._expand_list(bins)])
--- a/mkpkg
+++ b/mkpkg
@@ -158,6 +158,13 @@ class Functions:
                 continue
         return sorted(deps)
 
+    def mapfiles(self, src, dst, *bins, **kwargs):
+        append_suffix = kwargs.get('append_suffix', True)
+        suffix = self.pkg.vars['suffix'] if append_suffix else ''
+        return sorted(['{src}/{bin}={dst}/{bin}{suffix}'.format(
+                src=src, dst=dst, bin=b, suffix=suffix)
+                    for b in self._expand_list(bins)])
+
     def mapbins(self, src, dst, *bins):
         return sorted(['{src}/{bin}={dst}/{bin}{suffix}'.format(
                 src=src, dst=dst, bin=b, suffix=self.pkg.vars['suffix'])
--- a/README.rst
+++ b/README.rst
@@ -686,6 +686,13 @@ variable ``FUN``:
         is given, then all the ``bin`` passed will be prepended with this
         ``path``. ``bin``\ s can be passed as multiple arguments or as one
         list.
+``mapfiles(src, dst, file[, ...][, append_suffix=True])``
+        A very simple function that just returns a list with
+        ``{src}/{file}={dst}/{file}{VAR.suffix}`` for each ``file`` passed.
+        ``file``\ s can be passed as multiple arguments or as one list. A named
+        argument ``append_suffix`` can be passed at the end to control whether
+        ``VAR.suffix`` is appended to each destination file. ``append_suffix``
+        defaults to ``True`` if not given.
 ``mapbins(src, dst, bin[, ...])``
         A very simple function that just returns a list with
         ``{src}/{bin}={dst}/{bin}{VAR.suffix}`` for each ``bin`` passed.
--- a/README.rst
+++ b/README.rst
@@ -693,10 +693,6 @@ variable ``FUN``:
         argument ``append_suffix`` can be passed at the end to control whether
         ``VAR.suffix`` is appended to each destination file. ``append_suffix``
         defaults to ``True`` if not given.
-``mapbins(src, dst, bin[, ...])``
-        A very simple function that just returns a list with
-        ``{src}/{bin}={dst}/{bin}{VAR.suffix}`` for each ``bin`` passed.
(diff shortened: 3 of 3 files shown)


mkpkg
```diff
@@ -158,11 +158,15 @@ class Functions:
                 continue
         return sorted(deps)
 
-    def mapbins(self, src, dst, *bins):
+    def mapfiles(self, src, dst, *bins, **kwargs):
+        suffix = self.pkg.vars['suffix'] if 'append_suffix' in kwargs and kwargs['append_suffix'] else ''
```
gkotian: This line is quite long. It could be split up as:
```python
    suffix = vars_suffix if (
        'append_suffix' in kwargs and kwargs['append_suffix']
        ) else ''
```
but I'm not sure if that makes it better.

leandro-lucarella-sociomantic:
This should be based on v1.x.x really, because is a new feature.

RELEASE_NOTES.md
```diff
@@ -9,5 +9,5 @@ New Features
   - `FUN.autodeps()` now accepts a `path` to prepend to all passed binaries and
     can receive also the files as a single list instead of as multiple arguments.
 
-  - A new function 'FUN.mapbins()` was added to ease specifying binaries to
+  - A new function `FUN.mapbins()` was added to ease specifying binaries to
```
leandro-lucarella-sociomantic: Could you submit this in a separate PR against v1.8.x?

mkpkg
```diff
@@ -158,6 +158,12 @@ class Functions:
                 continue
         return sorted(deps)
 
+    def mapfiles(self, src, dst, *bins, **kwargs):
+        suffix = self.pkg.vars['suffix'] if 'append_suffix' in kwargs and kwargs['append_suffix'] else ''
```
leandro-lucarella-sociomantic: When the expression gets a bit complicated another option is just use a classical `if` statement:
```py
suffix = ""
if 'append_suffix' in kwargs and kwargs['append_suffix']:
    suffix = self.pkg.vars['suffix']
```

But this is probably a shorter and more clear alternative:

```py
suffix = kwargs.get('append_suffix', "")
```

:)

mkpkg
```diff
@@ -165,9 +165,7 @@ class Functions:
                     for b in self._expand_list(bins)])
 
     def mapbins(self, src, dst, *bins):
-        return sorted(['{src}/{bin}={dst}/{bin}{suffix}'.format(
-                src=src, dst=dst, bin=b, suffix=self.pkg.vars['suffix'])
-                    for b in self._expand_list(bins)])
+        return self.mapfiles(src, dst, *bins, append_suffix = True)
```
leandro-lucarella-sociomantic: Not a biggie, but in Python the convention is to not leave spaces around the `=` when passing keyword arguments.

leandro-lucarella-sociomantic (changes requested):
I'm not sure if it's worth keeping both really. Specially I don't like the asymmetry in the handling of the suffix in both `map` functions, and if we add the suffix handling to `mapbins` then both functions are identical, right?

I was assuming that binaries will always get the suffix, but then some users expressed the intention of not doing that, which would also make the existent of the specialization less justified.

I wonder.... :thinking:

RELEASE_NOTES.md
gkotian: https://github.com/sociomantic-tsunami/makd/pull/58

gkotian:
Thanks for the review, and for the tips on making the code more pythonic.

>I'm not sure if it's worth keeping both really. Specially I don't like the asymmetry in the handling of the suffix in both map functions, and if we add the suffix handling to mapbins then both functions are identical, right?

My main motivation for keeping both functions was to cause zero changes to any code that is already using `FUN.mapbins()`. But otherwise, I also don't think there is much value in keeping both functions.
Since makd's packaging feature is clearly marked as still being a WIP, I think it is reasonable to expect any users of `mapbins()` to make minor adjustments to their code.

I'll update the PR to remove `mapbins()` and only keep `mapfiles()`.

gkotian:
Updated with `FUN.mapbins()` renamed to `FUN.mapfiles()` and with the ability to control whether `VAR.suffix` is appended to the destination file name or not.

llucax:
> Updated with FUN.mapbins() renamed to FUN.mapfiles() and with the ability to control whether VAR.suffix is appended to the destination file name or not.

Oh! Sorry, i didn't mean to remove mapbins completely but just deprecate it and remove it in 2.0.0 to follow semver/Neptune.

llucax:
Also, please don't mix Bugfixes (typos), new features, deprecations and/or breaking changes in the same PR, all of those types of changes usually need to go to specific branches so they can't be all merged as part of the same PR:)

llucax:
So, to summarise: all bug fixes please submit against v1.8.x

The addition of mapfiles and deprecation of mapbins to v1.x.x (will target v1.9.x)

The removal of mapbins to 2.x.x.

Thanks!

gkotian:
>Oh! Sorry, i didn't mean to remove mapbins completely but just deprecate it and remove it in 2.0.0 to follow semver/Neptune.

How can I mark a function as deprecated in Python? There's one approach described in [this stackoverflow answer](http://stackoverflow.com/a/30253848), but I don't know if that's the recommended way.

gkotian:
>Also, please don't mix Bugfixes (typos), new features, deprecations and/or breaking changes in the same PR, all of those types of changes usually need to go to specific branches so they can't be all merged as part of the same PR:)

I guess you're referring to the `Fix typo` commit. That particular typo was introduced in 63f59dc9b4c048affadcd6f6b46f87c95c98c52e which afaict is not in any release. So including that fix in this PR (which targets `v1.x.x`) is correct I think. Can you confirm?

gkotian:
Updated to perform a sort of "soft deprecation" of `FUN.mapbins()`. This means that the release notes specifically states that `FUN.mapbins()` is deprecated and all mentions of it from the README are removed. However if someone still uses it, no deprecation warnings will be seen.

leandro-lucarella-sociomantic:
> How can I mark a function as deprecated in Python? There's one approach described in this stackoverflow answer, but I don't know if that's the recommended way.

I would keep it simple, just `warnf()` when the *deprecated* function is used, that's all.

leandro-lucarella-sociomantic:
> I guess you're referring to the Fix typo commit. That particular typo was introduced in 63f59dc which afaict is not in any release. So including that fix in this PR (which targets v1.x.x) is correct I think. Can you confirm?

You are right, disregard my message.

mkpkg
```diff
@@ -158,6 +158,13 @@ class Functions:
...
         return sorted(deps)
 
+    def mapfiles(self, src, dst, *bins, **kwargs):
+        append_suffix = kwargs.get('append_suffix', True)
+        suffix = self.pkg.vars['suffix'] if append_suffix else ''
+        return sorted(['{src}/{bin}={dst}/{bin}{suffix}'.format(
+                src=src, dst=dst, bin=b, suffix=suffix)
+                    for b in self._expand_list(bins)])
+
     def mapbins(self, src, dst, *bins):
```
leandro-lucarella-sociomantic, line 168: Here, just add `warnf("FUN.mapbins() is deprecated, please use FUN.mapfiles() instead')`.

leandro-lucarella-sociomantic (approved):
Once the warning is added, this looks good to me. In the meantime, can other people test this too?

@joseph-wakeling-sociomantic @nemanja-boric-sociomantic maybe? :)

leandro-lucarella-sociomantic:
(needs rebasing too, but better wait until #61 is merged for that, I guess)

gkotian:
Rebased over latest `v1.x.x` and added warning.

PR #58, opened 2017-03-10
[Trivial] Minor formatting fix

gkotian:


llucax (approved):

leandro-lucarella-sociomantic:
This should have gone to v1.7.x which is still supported. But being trivial documentation fixes I think I will skip it for v1.7.x which will also fall out of support soon.

My bad :(

PR #60, opened 2017-03-15
mkpkg: Add FUN.desc()

leandro-lucarella-sociomantic:
A very common pattern when creating packages for a multi-package project is to have a `pkg/defaults.py` file defining the common fields like a description.

The problem is, customizing this description in real package definition files where the defaults are imported is cryptic and complicated for users not familiar with Python.

This new functions makes it easier to customize general descriptions shared by several packages in multi-package projects.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -690,6 +690,25 @@ variable ``FUN``:
         A very simple function that just returns a list with
         ``{src}/{bin}={dst}/{bin}{VAR.suffix}`` for each ``bin`` passed.
         ``bin``\ s can be passed as multiple arguments or as one list.
+``desc(OPTS, [type[, prolog[, epilog]]])``
+        A simple function to customize ``OPTS['description']``. It can add an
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
I wanted to avoid having to pass `OPTS` but it proved to be a bit complicated, as inside the function we don't have access to the `OPTS` variable. There is a trick to do this, but it involves changing the way the `OPTS` and `VARS` variables are defined by the user, a breaking change. So I will delay that for v2, although I have the implementation ready, I'll wait to send the PR after this is merged so it can be built on top of this.

leandro-lucarella-sociomantic:
Again, I did some local testing, but test is super welcome!

@joseph-wakeling-sociomantic :upside_down_face:

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
In addition tested and LGTM

PR #69, opened 2017-03-25
mkpkg: Add ability to map entire directories

gkotian:
Currently, 'FUN.mapfiles()' returns an empty list on attempting to
simply map directories (i.e. with no file names passed into the
function). Fix this so that the function returns a simple mapping from
the given source directory to the given destination directory if no file
names are passed.

I'm targeting `v1.9.x` as I consider this a bugfix rather than a new feature. There is also no release notes update for the same reason.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,7 @@ also supported (see under `D2 support`_ for details).
 Versioning
 ----------
 
-MakD complies with `Netptune <https://github.com/sociomantic-tsunami/neptune>`_
+MakD complies with `Neptune <https://github.com/sociomantic-tsunami/neptune>`_
 for versioning.
 
 Support Guarantees
--- a/README.rst
+++ b/README.rst
@@ -686,13 +686,17 @@ variable ``FUN``:
         is given, then all the ``bin`` passed will be prepended with this
         ``path``. ``bin``\ s can be passed as multiple arguments or as one
         list.
-``mapfiles(src, dst, file[, ...][, append_suffix=True])``
+``mapfiles(src, dst, [file[, ...]][, append_suffix=True])``
         A very simple function that just returns a list with
         ``{src}/{file}={dst}/{file}{VAR.suffix}`` for each ``file`` passed.
-        ``file``\ s can be passed as multiple arguments or as one list. A named
-        argument ``append_suffix`` can be passed at the end to control whether
-        ``VAR.suffix`` is appended to each destination file. ``append_suffix``
-        defaults to ``True`` if not given.
+        ``file``\ s can be passed as multiple arguments or as one list. If no
+        ``file``\ s are passed, a simple ``{src}/={dst}/`` mapping is returned
+        which results in the directory structure under ``src`` (including all
+        contained files) being replicated as-is from the local filesystem into
+        ``dst`` within the package. A named argument ``append_suffix`` can be
+        passed at the end to control whether ``VAR.suffix`` is appended to each
+        destination file. ``append_suffix`` defaults to ``True`` if not given,
+        and is only applicable when at least one ``file`` is passed.
 ``desc(OPTS, [type[, prolog[, epilog]]])``
         A simple function to customize ``OPTS['description']``. It can add an
         optional ``type`` of package (will append `` (<type>)`` to the first
@@ -827,6 +831,7 @@ For convenience, here is a simple example:
 
         ARGS = FUN.mapfiles(VAR.bindir, '/usr/bin', bins)
         ARGS += FUN.mapfiles('.', '/etc', 'util.conf', append_suffix=False)
+        ARGS += FUN.mapfiles('doc', '/usr/share/doc')
 
 Suppose that the targets ``daemon`` and ``client`` build the binaries
 ``daemon``, ``admtool``, ``util1`` and ``client``, ``clitool`` respectively,
--- a/mkpkg
+++ b/mkpkg
@@ -159,11 +159,19 @@ class Functions:
         return sorted(deps)
 
     def mapfiles(self, src, dst, *bins, **kwargs):
-        append_suffix = kwargs.get('append_suffix', True)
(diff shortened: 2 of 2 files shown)


leandro-lucarella-sociomantic:
Mmm, if this is a bug fix I wonder if `mapbins` should get it too...

gkotian:
>Mmm, if this is a bug fix I wonder if mapbins should get it too...

Not worth the effort imo. `mapbins()` got deprecated not too long after it came into existence, so doesn't really have that many users I guess.

README.rst
```diff
@@ -686,13 +686,17 @@ variable ``FUN``:
...
-        ``VAR.suffix`` is appended to each destination file. ``append_suffix``
-        defaults to ``True`` if not given.
+        ``file``\ s can be passed as multiple arguments or as one list. If no
+        ``file``\ s are passed, a simple ``{src}/={dst}/`` mapping is returned
+        which results in the directory structure under ``src`` (including all
+        contained files) being replicated as-is from the local filesystem into
+        ``dst`` within the package. A named argument ``append_suffix`` can be
+        passed at the end to control whether ``VAR.suffix`` is appended to each
+        destination file. ``append_suffix`` defaults to ``True`` if not given,
+        and is only applicable when at least one ``file`` is passed.
```
leandro-lucarella-sociomantic, line 699: Mmm, I don't like that this function have so many different behaviours now based on the combination of options you pass. And honestly using `FUN.mapfiles('src', 'dst')` vs using `"src/=dst/"` makes little to no sense (or am I missing something here?). If the suffix were somehow handled at least (but I don't even know then if it should be added to the `dst` directory, but in that case is the same as `"src/=dst"+VAR.suffix+"/"`, which doesn't seems particularly complicated either).

leandro-lucarella-sociomantic (commented):
TBH I don't see a case for this... :confused:

README.rst
gkotian: >And honestly using FUN.mapfiles('src', 'dst') vs using "src/=dst/" makes little to no sense (or am I missing something here?)

I created a package definition file which initially had something like:
```
ARGS = FUN.mapfiles(VAR.bindir, '/usr/bin', app1, app2)
ARGS += FUN.mapfiles('configs', '/etc/configs', config1, config2) + [
    'helpers/=/etc/helpers'
]
```

In an attempt to make it more "symmetrical", I thought I could instead use:
```
ARGS = FUN.mapfiles(VAR.bindir, '/usr/bin', app1, app2)
ARGS += FUN.mapfiles('configs', '/etc/configs', config1, config2) 
ARGS += FUN.mapfiles('helpers', '/etc/helpers')
```

I figured that at best it would do the most intuitive thing - exactly replicate everything under the local `helpers` directory in the remote `/etc/helpers` directory; and at worst give me a build error on the third line saying something like "insufficient number of arguments given". But neither of these happened, and it returned an empty list instead. This, imo, is a bug.

I thought the best way to fix the bug is to provide what I consider the most intuitive behaviour when no file names are explicitly passed. So although technically there's no difference between using `FUN.mapfiles('src', 'dst')` and `src/=dst/`, I feel that the former way improves readability by providing a consistent look to the code.

leandro-lucarella-sociomantic: I'm sorry but I'm still not convinced :confused: 

For me adding a wrapper that does basically nothing is making things more obscure. At the end this is calling `fpm`, and I feel by abusing of `FUN.mapfiles()` we are making users of this wrapper distance themselves from the original format, that on itself has its own subtleties (like is not the same if you use a trailing `/` or not).

The `mapfiles()` function (or more correctly the original `mapbins()` function existed only to deal with the annoyance of having to include the suffix for a bunch of files, but when you remove the suffix from the scene, for me there is no place for this function. Maybe except (again) if you need to map a bunch of individual files, in which case it saves you from repetition (or more complex python list comprehensions). But if you don't need to add suffixes and all your files are in the same directory, unless there are other files in that directory that you don't want to include, I don't even see a case for `mapfiles()` as you can map the whole directory as `src/=dst`.

So again, `mapfiles()` was supposed to remove boilerplate, not just to add a layer of obscurity over a very clear and simple `src/=dst` string.

To finish I want to stress that I agree that the code looks more regular and nicer when using `mapfiles()` everywhere. Like is *graphically* nicer, but I think in terms of what is actually going on there, is way more obscure. So, how I am supposed to know that `FUN.mapfiles(VAR.bindir, '/usr/bin', app1, app2)` and `FUN.mapfiles('configs', '/etc/configs', config1, config2)` are adding suffixes and `FUN.mapfiles('helpers', '/etc/helpers')` does not (BTW, I really wonder if you intended to add suffixes to configuration files). So you have two almost identical pieces of code doing totally different things. That for me is a false (and dangerous) *readability improvement*.

gkotian: Submitted #78 as an alternative solution.

leandro-lucarella-sociomantic:
> Submitted #78 as an alternative solution.

Shall we close this one then?

gkotian:
>>Submitted #78 as an alternative solution.

>Shall we close this one then?

I don't know, it's your call really. According to me, this is a superior manner of addressing the problem. To borrow a term from dconf 2017, this solution is the "pit of success" model, wherein even if a programmer makes a mistake, things still just work. #78 otoh disallows the programmer from making the mistake. I'm ok with either approach.

llucax:
> I don't know, it's your call really. According to me, this is a superior manner of addressing the problem. To borrow a term from dconf 2017, this solution is the "pit of success" model, wherein even if a programmer makes a mistake, things still just work. #78 otoh disallows the programmer from making the mistake. I'm ok with either approach.

The problem here is making this a "pit of success" could also make the code for other people overly verbose, as [I explain](https://github.com/sociomantic-tsunami/makd/pull/80#issuecomment-299697937) in #80. In any case, I definitely think #80 is better than this approach, so I will close this one as wontfix and we'll see where #80 leads us.

Issue #59, opened 2017-03-14
Introduce standard `example` folder


mihails-strasuns-sociomantic:
Having detailed well-documented usage examples improves learning curve a lot so it is common for more mature projects to provide those. Within current makd layout we are supposed to put them to `test` folder but that is not very convenient - examples may be expected to fail for the very purpose of example or imply some environment setup that is not practical for testing purpose. It is also harder to find them when they are stored in same folder as tests.

Proposal is to define new `example` top-level folder in project directory and make makd compile its content in a way similar to `test` but without actually running resulting binaries (at least by default). Bonus points for modifying `make doc` target so that those example are available in dedicated documentation section.

leandro-lucarella-sociomantic:
It seems like a good idea. I think ideally examples should be ran too, and if it's supposed to fail check that it does fail. Maybe some sort of special metadata should be included, like which output is expected, or which exit code (including options to ignore those). So this really serves as both complete tests and documentation. But the distinction seems important enough (writing examples and tests have completely different goals even if they have a lot in common). Extending testing to be able to check for programs that are supposed to fail could be a good addition anyway.

Issue #41, opened 2017-01-31
Remove `-unittest` from default `test/` targets flags and run unittests in `test/` via `make unittest` instead


mihails-strasuns-sociomantic:
Current proposed solution
=====================

* Remove `-unittest` when compiling integrations tests (`test/`) because they complicate debugging.

* When getting the files to run `unittest`s for the `make unittest` target, also consider the integration tests files (in `test/`), so integrations test there still ran at least in some context.

This is a good trade-off to have integration tests still running but keeping clean runs for integration tests (also avoid a lot of duplication when running `unittest` blocks, which now  will be ran only once via the `make unittest` target).

Original issue description
====================

Originally I have pushed for keeping the flag so that internal sanity of test suite can also be validated. However practice has proven that this is almost never used. At the same time, running all unittests of all modules has also proven to complicate debugging a lot.

I propose to publicly shame me and remove that flag by default.

leandro-lucarella-sociomantic:
As long as you are shamed in public, I'm fine with it :trollface: 

Now being serious, I'm a bit concerned about user expectations here now. People might have started putting `unittest` blocks in `test/` and they'll expect for them to be tested too. So I think we need to go at least through a deprecation phase if we are going with this.

We could do:

1. Deprecation phase (v1.8.x)

   * Add a config option to enable the new behaviour (for example `TEST_WITHOUT_UNITTEST := 1`)
   * Add a warning when this new behaviour is not enabled explaining that the default will change in the future

2. Removal phase (v2.0.0)

   * Change default behaviour to `TEST_WITHOUT_UNITTEST := 1`
   * Optionally ignore the variable and warn the user if the variable is defined saying it's not used anymore (but maybe is good to leave this option, maybe for some cases enabling unittest blocks for integrations tests is desirable).

mathias-lang-sociomantic:
> People might have started putting unittest blocks in test/ and they'll expect for them to be tested too. So I think we need to go at least through a deprecation phase if we are going with this.

Note that for us, we can test this with:

> user:youknowho unittest path:test

It's also possible that a testsuite have unittests. What do we do in that case ?

leandro-lucarella-sociomantic:
> user:youknowho unittest path:test

I don't understand what this means.

> It's also possible that a testsuite have unittests. What do we do in that case ?

[...]

PR #77, opened 2017-04-24
Add new 'example' target

oraby:
I didn't find a tag for branch v2.x.x, so I wasn't sure if it was officially released yet. I've made the PR against v1.x.x. for now.

This PR adds new targets: `make examples` and `make examples-run`. The motivation is the following
1. Unit-tests are usually very focused on testing one specific aspect rather than giving a usage example.
2. Usage examples in documentation can get outdated and wouldn't be noticed.
3. At many times it's not convenient to browse the implementation code and understand the usage, especially that class implementations use inheritance and are spread over multiple modules.

So what this PR does is that it adds `make examples`. When executed, make will search for  `examples/` directory in the project root dir. The directory should contain  'real-life' examples of how code usage. CI tools can even optionally check whether example still compiles with each new PR.

If the general idea is accepted, then there are few small things to agree on:
1. Whether `examples` is a good target name or something else should be used (e.g `make samples`).
2. The `examples/` directory can contain non-`*.d` files (e.g config files). Should these be copied along to the output directory?
3. Should `test` target include `examples` target as well or should it be left to the user?

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -565,6 +565,40 @@ $O/test-%.stamp: $O/test-%
 	$(call exec,$< $(ITFLAGS),$<,run)
 	$Vtouch $@
 
+# examples rules
+##########################
+
+# examples are assumed to be standalone programs, so we recursively search
+# for files examples/*.d and assume that each file is a separate example
+# program. Files can exist in sub-directories under examples/ dir.
+# The sources list is filtered through the $(EXAMPLES_FILTER_OUT) variable 
+# contents (using the Make function $(filter-out)), so you can exclude an
+# example by adding the example file path (as an absolute path using $C) to
+# this variable.
+
+# The 'example' target builds all the examples.
+.PHONY: example
+example: $(patsubst $T/example/%.d,$O/example/%,\
+	$(filter-out $(EXAMPLE_FILTER_OUT),$(shell find $T/example/ -name "*.d")))
+
+# General rule to build an example program.
+$O/example/%: BUILD.d.depfile = $O/example/$*.mak
+$O/example/%: $T/example/%.d $G/build-d-flags
+	$(call build_d)
+	$Vdirname $(patsubst $O/example/%,$B/example/%,$@) | xargs mkdir -p
+	$Vcp $@ $(patsubst $O/example/%,$B/example/%,$@)
+
+# The 'example-run' target builds and runs all the examples.
+.PHONY: example-run
+example-run: $(patsubst $T/example/%.d,$O/example/%.run,\
+	$(filter-out $(EXAMPLE_FILTER_OUT),$(shell find $T/example/ -name "*.d")))
+
+# General rule to run an example program.
+# Don't touch a timestamp file so we would compile once but run every time.
+$O/example/%.run: $O/example/%
+	$(eval EXECUTABLE := $(patsubst $O/example/%.run,$B/example/%,$@))
+	$(call exec,$(EXECUTABLE),$(EXECUTABLE),run example)
+
 # Documentation rules
 ######################
 
--- a/README.rst
+++ b/README.rst
@@ -270,6 +270,8 @@ whole set of predefined targets are:
 * ``allunittest``
 * ``fastunittest``
 * ``integrationtest``
+* ``example``
+* ``example-run``
 * ``doc``
 * ``pkg``
 * ``graph-deps``
@@ -289,6 +291,12 @@ The ``fasttest`` target will only run the ``fastunittest`` target by default,
 but you can add more too by using the ``fasttest`` special variable.
 
 See the Testing_ section for more details.
+
+The ``example`` target will compile every ``.d`` file found under the
+``example/`` directory. The ``example-run`` target will compile and run
+all the found examples. An example can be skipped by adding the file to the
+``$(EXAMPLE_FILTER_OUT)`` variable. Check `Skipping tests`_ section for a
+similar example of filtering out files.
 
 The ``pkg`` target builds all packages defined in ``$P``, see Packaging_
 section for more details.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,10 @@
 New Features
 ============
 
+* `make example example-run`
+
+  Added new make targets 'example' and 'example-run'. The make targets will search for an `example/` directory in the root dir of the project. Each `*.d` file found will be compiled and copied into `build/last/bin/example/` directory. If `make example-run` is used, then the programs will be executed as well.
+
 * `VAR.fullname` `VAR.shortname`
 
   Added new variables, available in package definition files. `VAR.fullname` replaces the old `VAR.name` to make it more explicit, `VAR.shortname` contains only the package name, without the `VAR.suffix`.


leandro-lucarella-sociomantic:
This is fixing #59 essentially, so the commit should mention that.

> I didn't find a tag for branch v2.x.x, so I wasn't sure if it was officially released yet. I've made the PR against v1.x.x. for now.

Even when technically it could be breaking (if you are already using an `examples/` directory) I think is pretty unlikely that it does break, so I would include this in v1.10.0. @mihails-strasuns-sociomantic I don't know what do you think about it.

Since everything else is in singular, I would name the target and directory `example` instead, and I wonder if the directory should go to `doc/example` instead.

I will give a review to the code itself later today or maybe tomorrow.

mihails-strasuns-sociomantic:
Agreed that:
- it is a legit minor release for v1.x.x
- `example` is in style with existing stuff

Disagree about `doc/examples` though - it is still valid compilable code first from the build system PoV and only documentation material as a secondary purpose.

oraby:
> This is fixing #59 essentially, so the commit should mention that.

uh, I wasn't aware about that issue. I'll fix it late tonight.

> Since everything else is in singular, I would name the target and directory example instea

Cool, I'll update it then to `example`

> > and I wonder if the directory should go to doc/example instead.

> Disagree about doc/examples though - it is still valid compilable code first from the build system PoV and only documentation material as a secondary purpose.

I think I agree with MIhails on this one.

leandro-lucarella-sociomantic:
What do you mean secondary? Examples are primarily documentation! I don't feel too strong about `doc/example` but mainly because they get less visibility, but I think that's also because for now basically we never put much in `doc/` so nobody cares about it.

oraby:
> What do you mean secondary? Examples are primarily documentation! I don't feel too strong about doc/example but mainly because they get less visibility, but I think that's also because for now basically we never put much in doc/ so nobody cares about it.

I don't personally have a strong opinion about it either. However, since `makd` doesn't specify what the `doc/` directory at the project root directory should contain, so the user is free to use it for anything. A typical user case would be to store the automaticaly generated documentation there. In such case, an `example/` directory at that level might go unnoticed or might be not a convenient location that we however force on the user to use.

oraby:
Pushed a new commit. I'll update as necessary once we agree on the location of the examples directory.

Old head was at c76f4c8, new head is at 37c3f46: 
```diff
diff --git a/Makd.mak b/Makd.mak
index 15dd8d5..6e77346 100644
--- a/Makd.mak
+++ b/Makd.mak
@@ -576,27 +576,27 @@ $O/test-%.stamp: $O/test-%
 # example by adding the example file path (as an absolute path using $C) to
 # this variable.
 
-# The 'examples' target builds all the examples.
-.PHONY: examples
-examples: $(patsubst $T/examples/%.d,$O/examples/%,\
-	$(filter-out $(EXAMPLES_FILTER_OUT),$(shell find $T/examples/ -name "*.d")))
+# The 'example' target builds all the examples.
+.PHONY: example
+example: $(patsubst $T/example/%.d,$O/example/%,\
+	$(filter-out $(EXAMPLE_FILTER_OUT),$(shell find $T/example/ -name "*.d")))
 
 # General rule to build an example program.
-$O/examples/%: BUILD.d.depfile = $O/examples/$*.mak
-$O/examples/%: $T/examples/%.d $G/build-d-flags
+$O/example/%: BUILD.d.depfile = $O/example/$*.mak
+$O/example/%: $T/example/%.d $G/build-d-flags
 	$(call build_d)
-	$Vdirname $(patsubst $O/examples/%,$B/examples/%,$@) | xargs mkdir -p
-	$Vcp $@ $(patsubst $O/examples/%,$B/examples/%,$@)
+	$Vdirname $(patsubst $O/example/%,$B/example/%,$@) | xargs mkdir -p
+	$Vcp $@ $(patsubst $O/example/%,$B/example/%,$@)
 
-# The 'examples-run' target builds and runs all the examples.
-.PHONY: examples-run
-examples-run: $(patsubst $T/examples/%.d,$O/examples/%.run,\
-	$(filter-out $(EXAMPLES_FILTER_OUT),$(shell find $T/examples/ -name "*.d")))
+# The 'example-run' target builds and runs all the examples.
+.PHONY: example-run
+example-run: $(patsubst $T/example/%.d,$O/example/%.run,\
+	$(filter-out $(EXAMPLE_FILTER_OUT),$(shell find $T/example/ -name "*.d")))
 
 # General rule to run an example program.
 # Don't touch a timestamp file so we would compile once but run every time.
-$O/examples/%.run: $O/examples/%
-	$(eval EXECUTABLE := $(patsubst $O/examples/%.run,$B/examples/%,$@))
+$O/example/%.run: $O/example/%
+	$(eval EXECUTABLE := $(patsubst $O/example/%.run,$B/example/%,$@))
 	$(call exec,$(EXECUTABLE),$(EXECUTABLE),run example)
 
 # Documentation rules
diff --git a/README.rst b/README.rst
index 62b3343..12df3c8 100644
--- a/README.rst
+++ b/README.rst
@@ -270,8 +270,8 @@ whole set of predefined targets are:
 * ``allunittest``
 * ``fastunittest``
 * ``integrationtest``
-* ``examples``
-* ``examples-run``
+* ``example``
+* ``example-run``
 * ``doc``
 * ``pkg``
 * ``graph-deps``
@@ -292,10 +292,10 @@ but you can add more too by using the ``fasttest`` special variable.
 
 See the Testing_ section for more details.
 
-The ``examples`` target will compile every ``.d`` file found under the
-``examples/`` directory. The ``examples-run`` target will compile and run
+The ``example`` target will compile every ``.d`` file found under the
+``example/`` directory. The ``example-run`` target will compile and run
 all the found examples. An example can be skipped by adding the file to the
-``$(EXAMPLES_FILTER_OUT)`` variable. Check `Skipping tests`_ section for a
+``$(EXAMPLE_FILTER_OUT)`` variable. Check `Skipping tests`_ section for a
 similar example of filtering out files.
 
 The ``pkg`` target builds all packages defined in ``$P``, see Packaging_
diff --git a/RELEASE_NOTES.md b/RELEASE_NOTES.md
index 35707af..3dfdc03 100644
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,9 +1,9 @@
 New Features
 ============
 
-* `make examples examples-run`
+* `make example example-run`
 
-  Added new make targets 'examples' and 'examples-run'. The make targets will search for an `examples/` directory in the root dir of the project. Each `*.d` file found will be compiled and copied into `build/last/bin/examples/` directory. If `make examples-run` is used, then the programs will be run as well.
+  Added new make targets 'example' and 'example-run'. The make targets will search for an `example/` directory in the root dir of the project. Each `*.d` file found will be compiled and copied into `build/last/bin/example/` directory. If `make example-run` is used, then the programs will be executed as well.
 
 * `VAR.fullname` `VAR.shortname`
 ```

**Comment by mihails-strasuns-sociomantic** [comment]
> What do you mean secondary?

That it is still a valid compilable application as its primary purpose. If not for desire to improve its visibility which warrants own top-level folder, I'd put it into `src/example` instead as a second option, not to `doc/example`.

File: Makd.mak [hunk 1]
```diff
@@ -565,6 +565,40 @@ $O/test-%.stamp: $O/test-%
...
+# program. Files can exist in sub-directories under examples/ dir.
+# The sources list is filtered through the $(EXAMPLES_FILTER_OUT) variable 
+# contents (using the Make function $(filter-out)), so you can exclude an
+# example by adding the example file path (as an absolute path using $C) to
+# this variable.
+
+# The 'example' target builds all the examples.
+.PHONY: example
+example: $(patsubst $T/example/%.d,$O/example/%,\
+	$(filter-out $(EXAMPLE_FILTER_OUT),$(shell find $T/example/ -name "*.d")))
```
leandro-lucarella-sociomantic, line 582: `EXAMPLE_FILTER_OUT` should be documented on top too, the same as it is documented for `TEST_FILTER_OUT`.

leandro-lucarella-sociomantic, line 582: This `find` command is definitely incorrect, at least if we want to allow examples that consist on more than one file. See how `integrationtests` are done. My suggestion is to support both:

1. Top-level `<name>.d` files for simple one-file examples
2. More complex examples that need multiple files, that should be in `<name>/main.d`

(I would extend this to `integrationtests` too, since in most cases tests are only one file and the extra depth is just annoying)

Makd.mak
```diff
@@ -565,6 +565,40 @@ $O/test-%.stamp: $O/test-%
...
+
+# General rule to build an example program.
+$O/example/%: BUILD.d.depfile = $O/example/$*.mak
```
leandro-lucarella-sociomantic, line 585: I wouldn't use a subdirectory because otherwise it needs to be created (look for `setup_build_dir__`). On the other hand, if people want to run the examples manually more often, maybe is better to give them more visibility and put them in a dedicated directory in the build directory instead of the temporary one. We could use `$G/example`.

Makd.mak
```diff
@@ -565,6 +565,40 @@ $O/test-%.stamp: $O/test-%
...
+$O/example/%: $T/example/%.d $G/build-d-flags
+	$(call build_d)
+	$Vdirname $(patsubst $O/example/%,$B/example/%,$@) | xargs mkdir -p
+	$Vcp $@ $(patsubst $O/example/%,$B/example/%,$@)
```
leandro-lucarella-sociomantic, line 589: If you want to put them in `$B` just compile them there in the first place, but I don't like the idea of mixing examples with the real repos binaries, I prefer to use `E := $G/example` and store them there.

Makd.mak
```diff
@@ -565,6 +565,40 @@ $O/test-%.stamp: $O/test-%
...
+
+# The 'example' target builds all the examples.
+.PHONY: example
+example: $(patsubst $T/example/%.d,$O/example/%,\
+	$(filter-out $(EXAMPLE_FILTER_OUT),$(shell find $T/example/ -name "*.d")))
+
+# General rule to build an example program.
+$O/example/%: BUILD.d.depfile = $O/example/$*.mak
+$O/example/%: $T/example/%.d $G/build-d-flags
+	$(call build_d)
```
leandro-lucarella-sociomantic, line 587: I guess is probably a good idea to compile examples with unittests, otherwise unittests will never be run. And BTW, I wouldn't run examples unittests as part of the regular tests... (see #41).

Makd.mak
```diff
@@ -565,6 +565,40 @@ $O/test-%.stamp: $O/test-%
...
+# The 'example-run' target builds and runs all the examples.
+.PHONY: example-run
+example-run: $(patsubst $T/example/%.d,$O/example/%.run,\
+	$(filter-out $(EXAMPLE_FILTER_OUT),$(shell find $T/example/ -name "*.d")))
+
+# General rule to run an example program.
+# Don't touch a timestamp file so we would compile once but run every time.
+$O/example/%.run: $O/example/%
+	$(eval EXECUTABLE := $(patsubst $O/example/%.run,$B/example/%,$@))
+	$(call exec,$(EXECUTABLE),$(EXECUTABLE),run example)
```
leandro-lucarella-sociomantic, line 600: Looking at this now, I think we should probably not run the examples at all. I guess a lot of examples could take user input, or need arguments, it seems like too much trouble to cover all cases, and in reality I can see the use case for this more like:

1. User wants to know how to do X
2. User looks at `example/` and find a suitable example.
3. User compiles or copy the example code to play with it and run it as they are pleased

I think maybe compiling is enough. Thoughts @mihails-strasuns-sociomantic ?

README.rst
```diff
@@ -290,6 +292,12 @@ but you can add more too by using the ``fasttest`` special variable.
 
 See the Testing_ section for more details.
 
+The ``example`` target will compile every ``.d`` file found under the
```
leandro-lucarella-sociomantic, line 295: I think examples should get their own section in the documentation. I would only mention what is the target for in here and refer to the section for details, where the exclusions and a more general reasoning why this exists it is given.

leandro-lucarella-sociomantic (changes requested):
Several comments, some are mostly required, some are suggestions, some are discussion points.

Makd.mak
mihails-strasuns-sociomantic: Yes, my original intention was to ensure they compile and link but don't run anything. Various I/O and environment assumption are extremely common in examples.

Makd.mak
oraby: > 2. More complex examples that need multiple files, that should be in <name>/main.d

I was thinking about this, however I was thinking that maybe a directory will map several related sub-system files together. And my thinking was if one file is deemed to be imported and not stand-alone compiled then you'd add it to the `EXAMPLE_FILTER_OUT` example.

The best case I think would to check if the module has a main function, if so then we compile it, if not then it's being imported. Do you know if we can do this?

oraby: As far as I can see, the integration tests only supports `test/*/main.d`, right?. If we decided to go with your suggestions then it'll need to be  `test/*.d test/*/main.d`, that's correct?

oraby: > The best case I think would to check if the module has a main function, if so then we compile it, if not then it's being imported. Do you know if we can do this?

Do you know libdparse can help with this somehow?

README.rst
oraby: Good idea

Makd.mak
oraby: True, I was thinking at the same time that if examples are being run as part of unit-tests then you need to make sure that ti doesn't fail in run-time as well.

So my idea is that if an project wants to make sure that the examples still run correctly then it would do `test += example-run`.

I'm okay to remove it though, I just wanted to share my thought process about it.

Makd.mak
oraby: > I guess is probably a good idea to compile examples with unittests

I'll add it then as a dependency for the unittest target

> And BTW, I wouldn't run examples unittests as part of the regular tests... (see #41).

Is that a subtle hint to me to address that issue as well :)?
I'm happy to but I think it'll be a breaking change that needs to go against the next major.

Makd.mak
oraby: This just a temp directory for creating `.mak` file and putting timestamps there. They are copied later to `$B/example/`.
Due my poor Makfile skills, I didn't manage to compile it directory in `$B/example` (also without creating the `.mak` file there).
Spiritual and technical Makfile guidance is much appreciated.

Makd.mak
oraby: > If you want to put them in $B just compile them there in the first place

That was my original plan, but as I mentioned above, I failed to do it in the 'proper' way.

> I don't like the idea of mixing examples.  I prefer to use E := $G/example and store them there.

I put them there as they are some sort of binaries as well (e.g unlike `pkg`). But I agree yes that it's a bad idea to mix them with binaries. I'll implement your suggestion :+1: .

Makd.mak
leandro-lucarella-sociomantic: > As far as I can see, the integration tests only supports test/*/main.d, right?. If we decided to go with your suggestions then it'll need to be test/*.d test/*/main.d, that's correct?

Yes and yes.

> Do you know libdparse can help with this somehow?

I think is a lot of unnecessary complexity. If we want to allow arbitrary depth, we can achieve it via only compiling `main.d`s.

Makd.mak
leandro-lucarella-sociomantic: You should be able to infer it from both the `integrationtest` and `$B/%:` rules. I hope that's enlightening, because to go further I would have to go to the code myself and have a look.

Makd.mak
leandro-lucarella-sociomantic: Nope, I addressed that already :) I though it might help as an example.

Makd.mak
leandro-lucarella-sociomantic: I think for now it seems weak enough to be removed, we can add it in the future if we find that it would be useful to have such a rule.

leandro-lucarella-sociomantic:
What's the status of this, do you plan to resume the work on it @oraby ? I want to do a sprint to release v1.10.0 and I would like to include this in it.

hatem-oraby-sociomantic:
I'll try to give it a go this weekend

leandro-lucarella-sociomantic:
I'm taking over this one :D 

#90.

PR #78, opened 2017-04-25
Small bug fixes

gkotian:
This is an alternative to #69. It disallows `FUN.mapfiles()` from being called without any files.

### DIFF
--- a/mkpkg
+++ b/mkpkg
@@ -158,12 +158,12 @@ class Functions:
                 continue
         return sorted(deps)
 
-    def mapfiles(self, src, dst, *bins, **kwargs):
+    def mapfiles(self, src, dst, *files, **kwargs):
         append_suffix = kwargs.get('append_suffix', True)
         suffix = self.pkg.vars['suffix'] if append_suffix else ''
         return sorted(['{src}/{bin}={dst}/{bin}{suffix}'.format(
                 src=src, dst=dst, bin=b, suffix=suffix)
-                    for b in self._expand_list(bins)])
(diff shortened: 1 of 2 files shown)


mkpkg
```diff
@@ -159,6 +159,9 @@ class Functions:
         return sorted(deps)
 
     def mapfiles(self, src, dst, *files, **kwargs):
+        if not files:
+            die("'FUN.mapfiles()' expects at least one file to be passed")
+
```
leandro-lucarella-sociomantic: This is a breaking change so it should go to v2.x.x. But I don't understand the rationale for this... What's the point exactly?

leandro-lucarella-sociomantic:
It would be good to split this in 2 PR, all the uncontroversial bug fixes and the change in behaviour that is in essence a breaking change and should go to v2.x.x.

mkpkg
gkotian: `FUN.mapfiles()` returns `[]` if it is called as:
```
FUN.mapfiles('path/to/src', 'path/to/dst')
```
Since calling it this way does not show any errors or warnings, it could very easily lead to the programmer assuming that it will return `path/to/src/=path/to/dst/`, while in reality it returns an empty list.

#69 set out to fix exactly this, but by actually making the function return `path/to/src/=path/to/dst/`.

gkotian:
Removed the controversial commit from this PR, moving it to #80 instead.

llucax:
I renamed the PR to match the contents

Issue #74, opened 2017-04-07
Use integrationtest instead of test (because of D2)


leandro-lucarella-sociomantic:
Using `test` as a package name unnecessarily complicates testing in D2 as it clashes with the `test()` function. Because of this MakD should start accepting also `integrationtest/` as a location for integration tests. This is a bit longer and uglier but it should reduce considerably the chances of clashes.

If a project is using `test/` to store integration tests, a deprecation warning should be emitted.

The use of `test/` for testing should be completely removed in [v2.0.0](https://github.com/sociomantic-tsunami/makd/milestone/10).

leandro-lucarella-sociomantic:
Done now.

PR #79, opened 2017-04-26
Move all unittest processing to the unittest target

### LINKED ISSUE #41: Remove `-unittest` from default `test/` targets flags and run unittests in `test/` via `make unittest` instead

leandro-lucarella-sociomantic:
This commit moves the compilation and running of unittests to the
`unittest` target(s) and remove the flags `-unittest -debug=UnitTest
-version=UnitTest` when building integrationtests.

Fixes #41.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -260,7 +260,7 @@ PKGVERSION := $(shell echo $(VERSION) | \
 # whitespaces and each pattern can have one '%' that's used as a wildcard.
 # For more information refer to the documentation:
 # http://www.gnu.org/software/make/manual/make.html#Text-Functions
-TEST_FILTER_OUT := $C/$(SRC)/%/main.d
+TEST_FILTER_OUT :=
 
 
 # Functions
@@ -488,7 +488,8 @@ fasttest := fastunittest $(fasttest)
 test := unittest $(test)
 
 # Files to be tested in unittests, the user could potentially add more
-UNITTEST_FILES += $(call find_files,.d,,$C/$(SRC),$(TEST_FILTER_OUT))
+UNITTEST_FILES += $(call find_files,.d,,$C/$(SRC),$(TEST_FILTER_OUT)) \
+		$(call find_files,.d,,$C/$(INTEGRATIONTEST),$(TEST_FILTER_OUT))
 
 # Files to test when using fast or all unit tests
 $O/fastunittests.d: $(filter-out %_slowtest.d,$(UNITTEST_FILES))
@@ -516,8 +517,10 @@ endif
 $O/%unittests.d: $G/build-d-flags
 	$(call exec,printf 'module $(patsubst $O/%.d,%,$@);\n\
 		$(TEST_RUNNER_STRING)\n\
-		\n$(foreach f,$(filter %.d,$^),\
-		import $(subst /,.,$(patsubst $C/$(SRC)/%.d,%,$f));\n)' > \
+		\n$(foreach f,$(filter $C/$(SRC)/%.d,$^),\
+		import $(subst /,.,$(patsubst $C/$(SRC)/%.d,%,$f));\n)\
+		$(foreach f,$(filter $C/$(INTEGRATIONTEST)/%.d,$^),\
+		import $(subst /,.,$(patsubst $C/%.d,%,$f));\n)' > \
 			$@,,gen)
 
 # Configure dependencies files specific to each special unittests target
@@ -537,7 +540,7 @@ $O/%unittests.stamp: $O/%unittests
 			))),-p $p) \
 		$(foreach p,$(notdir $(shell \
 			find $T/$(SRC) -maxdepth 1 -mindepth 1 -type d \
-			)),-p $p.) $(UTFLAGS),$<,run)
+			)),-p $p.) -p $(INTEGRATIONTEST). $(UTFLAGS),$<,run)
 	$Vtouch $@
 
 # Integration tests rules
@@ -564,7 +567,7 @@ test += integrationtest
 # building any other binary but including unittests too.
 $O/test-%: BUILD.d.depfile = $O/test-$*.mak
 $O/test-%: $T/$(INTEGRATIONTEST)/%/main.d $G/build-d-flags
-	$(call build_d,-unittest -debug=UnitTest -version=UnitTest)
+	$(call build_d)
 
 # General rule to Run the test suite binaries
 $O/test-%.stamp: $O/test-%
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
@mathias-baumann-sociomantic I would assign you but you didn't accept the invitation yet :)

leandro-lucarella-sociomantic:
BTW, untested, please do test!

Makd.mak
```diff
@@ -479,7 +479,8 @@ fasttest := fastunittest $(fasttest)
 test := unittest $(test)
 
 # Files to be tested in unittests, the user could potentially add more
-UNITTEST_FILES += $(call find_files,.d,,$C/$(SRC),$(TEST_FILTER_OUT))
+UNITTEST_FILES += $(call find_files,.d,,$C/$(SRC),$(TEST_FILTER_OUT)) \
+		$(call find_files,.d,,$C/test,$(TEST_FILTER_OUT))
```
mihails-strasuns-sociomantic: This results in stuff like this in allunittests.d
```D
import .docker.ocean.test.unixlistener.main.d;
```
Maybe has something to do how `find_files` processes qualified paths?

mihails-strasuns-sociomantic (changes requested):

Makd.mak
leandro-lucarella-sociomantic: Mmm, weird. It shouldn't.

mihails-strasuns-sociomantic:
Ping @leandro-lucarella-sociomantic

leandro-lucarella-sociomantic:
I found the problem, but we have  another issue, the extra `main()`s.

For now for the program we get around this by just not building/running unittests in the `main.d` file, but for tests is normal that they only have a `main.d` so skipping those will mean no running unittests at all. I think the only sensible thing to do here is to stop filtering-out `main.d` files and just force projects to use something like:

```d
version (UnitTest) {}
else
int main(...)
{
    // ...
}
```

Ideas/opinions?

leandro-lucarella-sociomantic:
@gavin-norman-sociomantic FYI, reading the last comment might be enough.

leandro-lucarella-sociomantic:
If we go this route, I need to update this PR to fix docs/release notes to note:

* The project's `main.d` files will no longer be auto-filtered-out from unittests (if we'll require that for `integrationtest` I think it doesn't make sense to exclude any files from `make unittest`, it's unnatural.

* All `main()` files need to be versioned-out when they are compiled with `version=UnitTest`.

* New `$(INTEGRATIONTEST)` variable.

mihails-strasuns-sociomantic:
I don't think we should do anything too smart here. This will come in a new makd major release - migration notes explaining the change and asking to adjust own code in appropriate manner would be enough.

leandro-lucarella-sociomantic:
> I don't think we should do anything too smart here. This will come in a new makd major release - migration notes explaining the change and asking to adjust own code in appropriate manner would be enough.

This means you agree with the proposed changes?

leandro-lucarella-sociomantic:
ping

mihails-strasuns-sociomantic:
What do you mean by `INTEGRATIONTEST` variable?

Makd.mak
```diff
@@ -36,6 +36,9 @@ F ?= devel
 # use "." for the current directory)
 SRC = src
 
+# Directory where all the integration tests are
+INTEGRATIONTEST = test
```
leandro-lucarella-sociomantic: This

leandro-lucarella-sociomantic:
Just a way to ease the path to using `integrationtest` instead in the future. Not really related to this but it was good to make it easier to find all the places where `test/` was used.

mihails-strasuns-sociomantic:
I see. Yeah, I do agree with propose path in that case.

leandro-lucarella-sociomantic:
OK, I will update with the docs before I forget what I've done :D

leandro-lucarella-sociomantic:
> What do you mean by INTEGRATIONTEST variable?

While at this, I think I will backport this to v1.x.x, so projects can already try changing `test/` to `integrationtest/` to see if that really fixes our problems with D2 and see if it fixes #74.

If this is the case, we can change the default in v2.x.x more confidently!

leandro-lucarella-sociomantic:
Updated! Please give this a close look before merging, it was a bit too long ago since last time I touched it :)

leandro-lucarella-sociomantic:
`INTEGRATIONTEST` backported via #82 which also addresses the related #74.

leandro-lucarella-sociomantic:
Added a commit to fix #74 (change the default location for integration tests).

leandro-lucarella-sociomantic:
ping @mihails-strasuns-sociomantic

mihails-strasuns-sociomantic:
Will re-test after #74 is merged

leandro-lucarella-sociomantic:
You mean #82, right?

mihails-strasuns-sociomantic:
Yep

mihails-strasuns-sociomantic:
Rebase?

leandro-lucarella-sociomantic:
Mmm, no? It seems to be at the current v2.x.x.

leandro-lucarella-sociomantic:
Ah, maybe you mean merge, because the other changes were done to v1.x.x, I think I will merge after the release of v1, but I guess I could merge now if you prefer.

leandro-lucarella-sociomantic:
OK, updated on top of the merge.

leandro-lucarella-sociomantic:
Created a new commit with just the merge: https://github.com/sociomantic-tsunami/makd/pull/83

leandro-lucarella-sociomantic:
So this PR can get less noisy after the merge is merged :D

leandro-lucarella-sociomantic:
OK, now it's clean again.

mihails-strasuns-sociomantic (approved):
Looks good!

leandro-lucarella-sociomantic:
:tada:

PR #80, opened 2017-04-29
mkpkg: Die if no files are passed to mapfiles

gkotian:
Allowing no files to be passed results in 'mapfiles()' returning an empty list.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,6 +11,10 @@ Migration Instructions
 * The deprecated packaging function `FUN.mapbins()` was removed, use
   `FUN.mapfiles()` instead.
 
+* `FUN.mapfiles()` could previously be called with only the source and
+  destination directories (i.e. without any files). Passing at least one file to
+  the function is now mandatory.
+
 * The deprecated packaging variable `VAR.name` was removed, use `VAR.fullname`
   instead.
 
--- a/mkpkg
+++ b/mkpkg
@@ -159,6 +159,8 @@ class Functions:
         return sorted(deps)
 
     def mapfiles(self, src, dst, *bins, **kwargs):
+        if not bins:
+            die("'FUN.mapfiles()' expects at least one file to be passed")
         append_suffix = kwargs.get('append_suffix', True)
         suffix = self.pkg.vars['suffix'] if append_suffix else ''
         return sorted(['{src}/{bin}={dst}/{bin}{suffix}'.format(


llucax:
> Allowing no files to be passed results in 'mapfiles()' returning an empty list.

I think this is the right thing to do, you pass an empty list of files to map, you get an empty list of *mapped* files. This is what happens with every `map` function out there I know (like in [py](https://docs.python.org/2/library/functions.html#map) or [D](https://dlang.org/phobos/std_algorithm_iteration.html#.map)) and it's usually a good thing for generic code. Suppose a case like this:

```py
bins = []
if condition:
    bins.append('the-bin')
# More stuff, maybe more bins are conditionally added
files = mapfiles(src, dst, *bins)
```
Before this commit, it works as expected, after this commit you need to reimplement this as something like:
```py
bins = []
if condition:
    bins.append('the-bin')
# More stuff, maybe more bins are conditionally added
if bins:
    files = mapfiles(src, dst, *bins)
```
Which seems a bit silly to me.

I know there are other cases where not passing files could happen accidentally, like what I think is your use case:
```py
files = mapfiles(src, dst)
```
But still, I'm not sure if this change is really justified. Do you still think it's justified after this comment? Any other people have any other opinion?

gkotian:
Hmm, fair point. What do you think about replacing the `die()` with a `warnf()` instead?

llucax:
> Hmm, fair point. What do you think about replacing the die() with a warnf() instead?

I would be fine with that if there is a way to silence the warning. If an user using `files = mapfiles(src, dst, [])` will always get a warning, then that warning would be just annoying the hell out of them. Unfortunately I don't see a way to silence the warning that is not adding the `if` and if we are forcing the user to use that idiom (or we spam the hell out of them), then for me a warning doesn't make sense, it should be either an error or go silent.

For example, some C compilers will warn if you have this `if (a && b || c)` and suggest using parenthesis to declare intentions, so to silence the warning is enough to do `if ((a && b) || c)`. I find that a reasonable trade-off.

So if you find any way for the user to declare intentions and avoid the warning in a sane way, I'm up for it.

leandro-lucarella-sociomantic:
This is a good example of a warning well used :)

https://github.com/sociomantic-tsunami/makd/pull/82

leandro-lucarella-sociomantic:
Ping! I want to release v2.0.0 soonish, so if you want to include this there, it needs your attention soonish too :)

gkotian:
Hmm, I don't feel so strongly about this any more. Let's close this for now, and see if someone actually ever runs into the problem.

PR #82, opened 2017-05-22
Add a warning when using "test" for integration tests

leandro-lucarella-sociomantic:
Also introduce the variable INTEGRATIONTEST.

To improve compatibility with D2, the default location of the integration tests will change from `test` to `integrationtest`, so we need to warn the user about this.

The warning can be disabled by explicitly defining the `INTEGRATIONTEST` variable.

This is the v1.x.x part of #74.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -36,9 +36,6 @@ F ?= devel
 # use "." for the current directory)
 SRC = src
 
-# Directory where all the integration tests are
-INTEGRATIONTEST ?= test
-
 # Directory were this makefile is located (this must be done BEFORE including
 # any other Makefile)
 MAKD_PATH := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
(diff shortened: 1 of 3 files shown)


Makd.mak
```diff
@@ -36,6 +36,15 @@ F ?= devel
 # use "." for the current directory)
 SRC = src
 
+# Directory where all the integration tests are
+ifndef INTEGRATIONTEST
```
mihails-strasuns-sociomantic: I tried defining `INTEGRATIONTEST:=test` in `Config.mak` but message keeps being printed

mihails-strasuns-sociomantic (changes requested):

Makd.mak
leandro-lucarella-sociomantic: Right! Currently it is a bit messy what can be overridden by `Config.mak`, I should probably clean that up at some point.

leandro-lucarella-sociomantic:
Updated.

mihails-strasuns-sociomantic:
Still doesn't work for me with `Config.mak`

leandro-lucarella-sociomantic:
That's weird, it works for me:
```
$ git fetch https://github.com/leandro-lucarella-sociomantic/makd.git deprecate-74
From https://github.com/leandro-lucarella-sociomantic/makd
 * branch            deprecate-74 -> FETCH_HEAD

$ git checkout -b test FETCH_HEAD

$ make SRC=. -f Makd.mak 
MakD Warning: The default location of integration tests (defined by $(INTEGRATIONTEST) and 'test' by default now) will change to 'integrationtest' in v2.0.0. If you want to avoid this warning, please define it explicitly in your Config.mak.
make: Nothing to be done for 'all'.

$ echo 'INTEGRATIONTEST := lala' > Config.mak

$ make SRC=. -f Makd.mak 
make: Nothing to be done for 'all'.
```

leandro-lucarella-sociomantic:
Ah, I think I pushed the latest changes after you tested, there was a pending issue :D

mihails-strasuns-sociomantic:
Yep, works now!

mihails-strasuns-sociomantic (approved):

PR #84, opened 2017-06-28
pkg: Add ability to use options without values

leandro-lucarella-sociomantic:
When an option has `True` as value, `mkpkg` will just pass `--key` (or
`-k`) to `fpm`, without any actual value. This allows using `fpm`
flags that don't take any value (on/off switches).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -642,10 +642,12 @@ These files are expected to be Python scripts defining two variables:
         a fpm_ command-line option. If the key is only one character (for
         example ``c``), it will be passed as ``-<key><value>`` and if it's
         more, it will be passed as ``--<key>=<value>`` (``_`` characters in the
-        key will be replaced by ``-`` for convenience). The ``<value>`` can be a
-        string or an array of strings. In the latter case, the key is used as fpm_
-        flag for each item in ``<value>``. No validation is performed over the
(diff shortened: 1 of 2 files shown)


nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
Thanks!

leandro-lucarella-sociomantic:
As a pseudo-workaround for older versions, `--default=--some-fpm-arg` or `--override=--some-fpm-arg` can be used when calling `mkpkg`.

leandro-lucarella-sociomantic:
The former will be used only if the user doesn't define the flag and the later will override any user-specified option (so in this case that you can't really specify it in the definition file, it would be the same).

nemanja-boric-sociomantic:
`override` and `default` are actually makd options, not fpm ones, right?

leandro-lucarella-sociomantic:
Exactly. The values you pass to those `mkpkg` options are passed to `fpm`.

nemanja-boric-sociomantic:
Thanks!

leandro-lucarella-sociomantic:
so `mkpkg -D--fpm-option` would call (at some point and in some way) to `fpm --fpm-option`.

PR #85, opened 2017-06-28
mkpkg: Use dpkg-shlibdeps in autodeps()

### LINKED ISSUE #76: FUN.autodeps produce NoOp dependencies
In one of our applications, I use `FUN.autodeps`.
Recently, as I was updating it on a server, I realized it failed to update `libdrizzle-redux`.
The old `libdrizzle-redux` used to compile the package was `v5.6.0`, but the server only had `v5.4.1`.
The newly deployed app required `v6.0.0` (hence a different soname, but at that time the package had the same name), and as a result failed to start because `apt-get` did not update the dependency.

I did a bit of investigation. First the output of `autodeps` is the following:
```
['libbz2-1.0', 'libc6', 'libdrizzle-redux', 'libgcc1', 'libgcrypt20', 'libgeoip1', 'libglib2.0-0', 'libgpg-error0', 'liblzo2-2', 'libmcrypt4', 'libpcre3', 'libssl1.0.0', 'libstdc++6', 'libtokyocabinet9', 'zlib1g']
```

When I run `dpkg -I` on the resulting package, I get:
```
[...]
 Depends: libbz2-1.0, libc6, libdrizzle-redux, libgcc1, libgcrypt20, libgeoip1, libglib2.0-0, libgpg-error0, liblzo2-2, libmcrypt4, libpcre3, libssl1.0.0, libstdc++6, libtokyocabinet9, zlib1g, libbz2-1.0, libc6, libdrizzle-redux, libgcc1, libgcrypt20, libgeoip1, libglib2.0-0, libgpg-error0, liblzo2-2, libmcrypt4, libpcre3, libssl1.0.0, libstdc++6, libtokyocabinet9, zlib1g, libbz2-1.0, libc6, libdrizzle-redux, libgcc1, libgcrypt20, libgeoip1, libglib2.0-0, libgpg-error0, liblzo2-2, libmcrypt4, libpcre3, libssl1.0.0, libstdc++6, libtokyocabinet9, zlib1g
[...]
```

As you can notice, the package includes 3 binaries, which have the same dependencies, so dependencies appear 3 times in the output.

When I run `apt-cache showpkg mypackage`, I get the following:
```
Reverse Depends: 
Dependencies: 
2.3.1-xenial - libc6 (0 (null)) libdrizzle-redux (0 (null)) libgcc1 (0 (null)) libgcrypt20 (0 (null)) libgeoip1 (0 (null)) libglib2.0-0 (0 (null)) libgpg-error0 (0 (null)) liblzo2-2 (0 (null)) libmcrypt4 (0 (null)) libpcre3 (0 (null)) libssl1.0.0 (0 (null)) libstdc++6 (0 (null)) zlib1g (0 (null)) libc6 (0 (null)) libdrizzle-redux (0 (null)) libgcc1 (0 (null)) libgcrypt20 (0 (null)) libgeoip1 (0 (null)) libglib2.0-0 (0 (null)) libgpg-error0 (0 (null)) liblzo2-2 (0 (null)) libmcrypt4 (0 (null)) libpcre3 (0 (null)) libssl1.0.0 (0 (null)) libstdc++6 (0 (null)) zlib1g (0 (null)) libc6 (0 (null)) libdrizzle-redux (0 (null)) libgcc1 (0 (null)) libgcrypt20 (0 (null)) libgeoip1 (0 (null)) libglib2.0-0 (0 (null)) libgpg-error0 (0 (null)) liblzo2-2 (0 (null)) libmcrypt4 (0 (null)) libpcre3 (0 (null)) libssl1.0.0 (0 (null)) libstdc++6 (0 (null)) zlib1g (0 (null)) 
```

The man page of `apt-cache showpkg` tell me to read the source code (!), so I did, and here's the [relevant section](https://github.com/Debian/apt/blob/2a90aa7a064047fb1c8783b31720cd345018ca4a/apt-pkg/pkgcache.h#L156). In short, `0` means the dependency is a no-op, and there's no associated version.

TL;DR: `FUN.autodeps` only ensures the package is installed based on the package name, but does not store version information, which can result in failures and manual dependency upgrade.

leandro-lucarella-sociomantic:
The naive approach of just listing packages using `ldd` + `dpkg -S`
doesn't work that well, because by not giving version restrictions,
upgrades are usually not handled properly, as some minimal version
needed by a new package might never be picked by `apt`.

This should fix #76.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -52,10 +52,12 @@ PKG := $T/pkg
 
 # Directory where all the integration tests are
 ifndef INTEGRATIONTEST
+ifeq ($(if $(wildcard test/*/main.d),yes),yes)
 __dummy_integrationtest_warning := $(shell echo "MakD Warning: The default \
 	location of integration tests (defined by \$$(INTEGRATIONTEST) and 'test' \
 	by default now) will change to 'integrationtest' in v2.0.0. If you want to \
 	avoid this warning, please define it explicitly in your Config.mak." >&2)
+endif
 endif
 INTEGRATIONTEST ?= test
 
--- a/README.rst
+++ b/README.rst
@@ -731,14 +731,18 @@ passed to the ``mkpkg`` util. By default Makd pass the following variables:
 variable ``FUN``:
 
 ``autodeps(bin[, ...][, path=''])``
-        returns a sorted ``list()`` of packages ``bin`` depends on based on the
-        outcome of running the ``ldd`` utility and searching to which packages
-        the libraries is linked belong to using ``dpkg``. You can specify
-        multiple binaries to get a list of dependencies for all of them. This
-        function is tightly coupled to Debian packages for now. If a ``path``
-        is given, then all the ``bin`` passed will be prepended with this
-        ``path``. ``bin``\ s can be passed as multiple arguments or as one
-        list.
+        returns a sorted ``list()`` of dependencies (packages with versions)
+        ``bin`` depends on based on the outcome of running the
+        ``dpkg-shlibdeps`` utility in a minimal bogus debian package
+        environment. This tools comes in the ``dpkg-dev`` package, so make sure
+        you have it installed if you use this function. You can specify
+        multiple binaries to get a list of dependencies for all of them, and if
+        a single package provides multiple binaries you should do **one
+        single** ``autodeps()`` call for all of them to avoid getting
+        duplicates. This function is tightly coupled to Debian packages for
+        now. If a ``path`` is given, then all the ``bin`` passed will be
+        prepended with this ``path``. ``bin``\ s can be passed as multiple
+        arguments or as one list.
 ``mapfiles(src, dst, file[, ...][, append_suffix=True])``
         A very simple function that just returns a list with
         ``{src}/{file}={dst}/{file}{VAR.suffix}`` for each ``file`` passed.
--- a/mkpkg
+++ b/mkpkg
@@ -89,34 +89,21 @@ def parse_args():
     return parser, args
 
 
+class Dependency:
+    def __init__(self, pkg, version=''):
+        self.pkg = pkg
+        self.version = version
+    def __str__(self):
+        version = self.version
+        if version:
+            version = ' ' + version
+        return self.pkg + version
+
 class Functions:
-
-    # lines: "\t<name> => <path> (<address>)"
-    #        "\t<name> => not found"
-    ldd_re = re.compile(r'^\s+(\S+) => (.*?)(?: \(.*\))?$', re.MULTILINE)
-
-    # lines: "<package>:<arch>: <file>"
-    dpkg_re = re.compile(r'^([^:]+)(?::[^:]+)?: .*$', re.MULTILINE)
-
-    autodeps_exclusions = frozenset(r'''
-            linux-vdso.so.1
-        '''.split())
 
     def __init__(self, args, pkg):
         self.args = args
         self.pkg = pkg
-
-    def _parse_cmd(self, cmd, *args):
-        args1 = cmd.split()
-        cmd = args1.pop(0)
-        regex = getattr(self, cmd + '_re')
-        cmd = [getattr(self.args, cmd + '_path')]
-        cmd.extend(args1)
-        cmd.extend(args)
-        debugf("Running command: {}", cmd)
-        output = subprocess.check_output(cmd).decode(sys.stdin.encoding)
-        debugf("Output: {}", output)
-        return regex.findall(output)
 
     def _expand_list(self, files, path=''):
         if path and not path.endswith('/'):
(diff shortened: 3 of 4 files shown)


leandro-lucarella-sociomantic:
Not only reviews, but also testing is welcome.

leandro-lucarella-sociomantic:
The repetition problem for user-provided dependencies that also appear in the `autodeps()` is still not handled. The code is fairly prepared to do so, though.

My plan is to somehow see if the user provides any `pkg` that is also present in `autodeps()` and override it with what the user specified, but it's not easy because for far there is no way for `autodeps()` to see if there are user-dependencies too.

One way to do this would be to add some sort of `extra_deps` parameter to `autodeps()` in which one can specify overrides. In this case users should provide all depdencies via `autodeps()`. Another approach would be to add another function like `overridedeps()`, but it might be a bit weaker because what overrides what would be determined by ther order in which dependencies are passed to this function.

I'm still not sure how `dpkg` handles duplicates, so maybe duplicates are not really an issue. If you can test with real-life cases, it would be very helpful!

leandro-lucarella-sociomantic:
Can someone give this some testing?

mihails-strasuns-sociomantic:
Tried with dhtnode-utils, got this version string from `apt-cache showpkg dhtnode-utils`:

```
1.10.0+3~dirty.20170818115221~934ae79-xenial - libc6 (2 2.14) libglib2.0-0 (2 2.12.0) liblzo2-2 (0 (null)) 
```

Looks better than the original case but there is still one null version. Expected?

leandro-lucarella-sociomantic:
I've been looking to other packages dependencies, and I think it seems fine. Other (official ubuntu) packages depending on liblzo2-2 also have null as a version. I think version are not null only when dpkg can find symbols that are in one version but not another one or something like that. Usually also all non-binary packages get version null. I think this is not magic, it won't be able to guess all dependencies with the exact versions, sometimes the developer will have to put the right dependencies when needed.

leandro-lucarella-sociomantic:
@mathias-lang-sociomantic it would be nice to have your feedback here, since you opened the original issue, but I think if we don't get feedback in one or two days we should merge.

mathias-lang-sociomantic:
```
geod24@komputer ~/projects/aquatic (git)-[v2.x.x] % make pkg                                                                                               MakD Warning: The default location of integration tests (defined by $(INTEGRATIONTEST) and 'test' by default now) will change to 'integrationtest' in v2.0.0. If you want to avoid this warning, please define it explicitly in your Config.mak.
   mkpkg pkg/aquatic.pkg
Error: can't load package definition file '/home/geod24/teams/mineobviously/aquatic/pkg/aquatic.pkg'

Traceback (most recent call last):
  File "/home/geod24/teams/stillmine/aquatic/submodules/makd/mkpkg", line 142, in dpkg_autodeps
    output = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
  File "/usr/lib/python3.5/subprocess.py", line 626, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.5/subprocess.py", line 708, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['dpkg-shlibdeps', '/home/geod24/teams/stillmine/aquatic/build/devel/bin/aquatic-stream']' returned non-zero exit status 2

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/geod24/teams/stillmine/aquatic/pkg/aquatic.pkg", line 13, in <module>
    OPTS['depends'] = FUN.autodeps(bin_path) + FUN.autodeps(iterate_path) + FUN.autodeps(mapview_path)
  File "/home/geod24/teams/stillmine/aquatic/submodules/makd/mkpkg", line 170, in autodeps
    deps = self.dpkg_autodeps(self._expand_list(fnames, path))
  File "/home/geod24/teams/stillmine/aquatic/submodules/makd/mkpkg", line 146, in dpkg_autodeps
    ' '.join(cmd), e.returcode, e.output)
AttributeError: 'CalledProcessError' object has no attribute 'returcode'
submodules/makd/Makd.mak:480: recipe for target '/home/geod24/teams/stillmine/aquatic/build/devel/tmp/pkg-aquatic.stamp' failed
make: *** [/home/geod24/teams/stillmine/aquatic/build/devel/tmp/pkg-aquatic.stamp] Error 1
```

leandro-lucarella-sociomantic:
This is one to make all those dynamic languages haters very happy... `returcode` -> `returncode` :snake:

leandro-lucarella-sociomantic:
Updated. Still I don't know why the command is returning 2.

leandro-lucarella-sociomantic:
The `dpkg-dev` package is needed (which I think it's undocumented), but I don't think that's a "command not found" error.

mathias-lang-sociomantic:
I have `dpkg-dev` installed

leandro-lucarella-sociomantic:
I added the dependency in the docs now. Can you see if now you get some more output about the original failure?

mathias-lang-sociomantic:
```
Warning: Looking for dependencies, command `dpkg-shlibdeps /path/to/binary` returned 2: b'dpkg-shlibdeps: error: no dependency information found for /usr/lib/libdrizzle-redux6.so.13 (used by /path/to/project)\nHint: check if the library actually comes from a package.\n'
Error: can't load package definition file 'project.pkg'

Traceback (most recent call last):
  File "project.pkg", line 13, in <module>
    OPTS['depends'] = FUN.autodeps(bin_path) + FUN.autodeps(iterate_path) + FUN.autodeps(mapview_path)
  File "submodules/makd/mkpkg", line 171, in autodeps
    return sorted(set(str(d) for d in deps))
TypeError: 'NoneType' object is not iterable
submodules/makd/Makd.mak:480: recipe for target 'build/devel/tmp/pkg-project.stamp' failed
make: *** [build/devel/tmp/pkg-project.stamp] Error 1
make pkg  21.64s user 2.59s system 94% cpu 25.754 total
```

mathias-lang-sociomantic:
I purged and reinstalled `libdrizzle-redux6`, and it does come from the package. The tool also shows a couple of warnings (e.g. I link with tokyocabinet but does not use it) which might or might not be helpful (there's a warning about `ld-linux`...)

leandro-lucarella-sociomantic:
> I purged and reinstalled libdrizzle-redux6, and it does come from the package. The tool also shows a couple of warnings (e.g. I link with tokyocabinet but does not use it) which might or might not be helpful (there's a warning about ld-linux...)

Mmm, damn, I wonder if the real dpkg tools are adding some dependency info somewhere that we are not producing by using `fpm`. I have the (sad) feeling we'll eventually have to switch to use the proper dpkg tools...

leandro-lucarella-sociomantic:
I accept suggestions on how to move forward...

mihails-strasuns-sociomantic:
I have no idea other than reporting the problem to fpm maintainers and asking if it has some arcane CLI flag to fix it.

mathias-lang-sociomantic:
Something like :

>     --deb-shlibs SHLIBS           (deb only) Include control/shlibs content. This flag expects a string that is used as the contents of the shlibs file. See the following url for a description of this file and its format: http://www.debian.org/doc/debian-policy/ch-sharedlibs.html#s-shlibs

? ;)

Libdrizzle-redux will have to provide a shlib file. As well as other libraries we package.
However we should look into *not* hard-erroring on those situations, as they can arise from packages not under our control.

leandro-lucarella-sociomantic:
More info also in https://www.debian.org/doc/debian-policy/ch-sharedlibs.html

leandro-lucarella-sociomantic:
It is a bit strange that is some sort of parallel dependency system... :unamused:

leandro-lucarella-sociomantic:
> However we should look into not hard-erroring on those situations, as they can arise from packages not under our control.

This would be extremely difficult, as `dpkg-shlibdeps` just fails and doesn't give us any dependencies.

leandro-lucarella-sociomantic:
There seems to be a `--ignore-missing-info`

leandro-lucarella-sociomantic:
Updated with the `--ignore-missing-info` flag, can you test again? My fear now is that no dependencies are generated at all for libs without info.

mihails-strasuns-sociomantic:
Ping @mathias-lang-sociomantic

mihails-strasuns-sociomantic:
Ping @mathias-lang-sociomantic

mathias-lang-sociomantic:
> Updated with the `--ignore-missing-info` flag, can you test again? My fear now is that no dependencies are generated at all for libs without info.

Yay it just skips over... Also the output is not great:
>  Depends: libc6 (>= 2.14), libgcrypt20 (>= 1.6.1), libgeoip1, libglib2.0-0 (>= 2.12.0), libgpg-error0 (>= 1.14), liblzo2-2, libmcrypt4, [Redundant...]

[I'll fix libdrizzle-redux (or not)](https://github.com/sociomantic-tsunami/libdrizzle-redux/issues/246) and look into why those libs don't have infos...

mathias-lang-sociomantic:
:sadface:

leandro-lucarella-sociomantic:
Oh, crap, this is because I removed the v1.x.x branch.

leandro-lucarella-sociomantic:
I'll have to do it against v2.x.x eventually

leandro-lucarella-sociomantic:
Duplicate of #116

PR #86, opened 2017-06-28
test: Add exception for importing x.y.package

### LINKED ISSUE #73: D2: Handle packages when building allunittests
When calling `make test`, MakD will generate a file `build/$F/tmp/allunittests.d` importing all files in a project and all dependencies.

In D2, if a file is called `package`, it can act as a way to import the package.
E.g. the file `src/altered/package.d` can have `module altered;` as ModuleDeclaration.

However, since MakD seems to rely on the file <-> ModuleDeclaration association, this use case will generate:

```
module allunittests;
 import ocean.core.UnitTestRunner;

 import altered.package;
```

Which will result in a compilation failure:
```
altereD/build/devel/tmp/allunittests.d(5): Error: identifier expected following package
altereD/build/devel/tmp/allunittests.d(5): Error: ';' expected
```

leandro-lucarella-sociomantic:
When generating the `allunitests.d` file, for every file an `import` is
generated based on the file name, but in D2 the file `pkg/package.d` is
also allowed, but is imported when `pkg` is imported

Since using `package` as a file/import name is invalid in D1, is safe to
make this exception regardless of the D version being used.

Fixes #73.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -415,6 +415,16 @@ define build_d
 	$(POST_BUILD_D) $3
 endef
 
+# Converts a file path to a D module name. It converts correctly
+# a `pkg/package.d` to just `pkg`.
+# This function accepts 2 arguments:
+# $1 is the file path to convert (required)
+# $2 is the base path of the file (that is not part of the module
+#    specification). By default $C/$(SRC).
+file2module = $(subst /,.,$(patsubst %/package,%,$(patsubst \
+		$(if $2,$2,$C/$(SRC))/%.d,%,$1)))
+
+
 # Overridden and default flags
 ###############################
 
@@ -540,8 +550,7 @@ $O/%unittests.d: $G/build-d-flags
 	$(call exec,printf 'module $(patsubst $O/%.d,%,$@);\n\
 		$(TEST_RUNNER_STRING)\n\
 		\n$(foreach f,$(filter %.d,$^),\
-		import $(subst /,.,$(patsubst $C/$(SRC)/%.d,%,$f));\n)' > \
-			$@,,gen)
+		import $(call file2module,$f);\n)' > $@,,gen)
 
 # Configure dependencies files specific to each special unittests target
 $O/fastunittests: BUILD.d.depfile := $O/fastunittests.mak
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
Also please test!

leandro-lucarella-sociomantic:
ping!

Makd.mak
```diff
@@ -522,8 +522,8 @@ $O/%unittests.d: $G/build-d-flags
 	$(call exec,printf 'module $(patsubst $O/%.d,%,$@);\n\
 		$(TEST_RUNNER_STRING)\n\
 		\n$(foreach f,$(filter %.d,$^),\
-		import $(subst /,.,$(patsubst $C/$(SRC)/%.d,%,$f));\n)' > \
-			$@,,gen)
+		import $(subst /,.,$(patsubst %/package,%,$(patsubst \
+				$C/$(SRC)/%.d,%,$f)));\n) > $@,,gen)
```
mathias-lang-sociomantic: > /bin/sh: 1: Syntax error: Unterminated quoted string

leandro-lucarella-sociomantic:
Maybe now?

mathias-lang-sociomantic:
Almost there. I added a `module altered; unittest { assert(0); }` in `src/altered/package.d`, everything compiled and run fine...
Turns out `allunittests` gets called as follows:
```
/home/geod24/tsunami/altereD/build/devel/tmp/allunittests  -v -s  -p altered.
```

And the package unittests don't get run. If called with:
```
/home/geod24/tsunami/altereD/build/devel/tmp/allunittests  -v -s  -p altered
```

That is, without the trailing dot, they get run as expected.

leandro-lucarella-sociomantic:
Mmm, that's strange, I'll check.

leandro-lucarella-sociomantic:
Ohhh, this is tricky... The problem is to build the `-p` arguments, I'm just checking the top-level packages/modules in `$(SRC)`. Directories are considered packages, and thus a `.` is appended at the end to make sure a naïve `pkglala` matches a `pkg` package (the unit test runner matches only substrings). If we want to detect `pkg/package.d` properly as a module, then a lot of logic should be added to something that now is pretty simple and it is already somewhat flakey.

Anyway, I think I just came up with a simple-ish way to fix it. Let's see.

leandro-lucarella-sociomantic:
OK, this time I even did some testing and it seems to be fine. I hope is the last try :)

leandro-lucarella-sociomantic:
Some final testing?

mathias-lang-sociomantic:
Will do on Monday (I'm on vacations :D)

mathias-lang-sociomantic:
Grey

leandro-lucarella-sociomantic:
Greening...

mihails-strasuns-sociomantic:
Tested it quickly by adding `ocean/package.d`. It kind of works but does so by providing `-p ocean -p ocean.` simultaneously which kind of kills the original point of using a dot in the first place. Now if there is top-level `project/package.d`, all tests from `projectzzzz` will run too.

I don't see how it can be fixed reliably - maybe better to instead focus on simplicity and just always supply package name without dot despite possible false positives?

mathias-lang-sociomantic:
It sounds to me that the problem is that we simply do an `strcmp` on the name, and the dot was a workaround for it. Is there a single case where we provide `xxx` and want `xxx*` to be tested ?

llucax:
> It sounds to me that the problem is that we simply do an strcmp on the name, and the dot was a workaround for it. Is there a single case where we provide xxx and want xxx* to be tested ?

Probably not, and I agree that the reliable solution must be implemented in ocean's unit test runner.

leandro-lucarella-sociomantic:
See https://github.com/sociomantic-tsunami/ocean/pull/221. I think in the future we should use that, but we need a temporary solution. I would go for what it is already in this PR, since it should be temporary anyway, but I could also only include `-p ocean`.

leandro-lucarella-sociomantic:
OTOH, I guess we'll need to support older oceans for quite a while, so maybe as a transition we need to actually detect if the unit test program supports `-m` and use it (or not) based on that...

mihails-strasuns-sociomantic:
Commented on the matter in ocean PR

leandro-lucarella-sociomantic:
I suggest to go forward with this PR as it is so we don't block the release, as the ocean PR could take a while and there is a chance that this can't be used immediately anyway.

This PR implements the more complicated approach, so adapting it to any solution in ocean would always be a simplification.

mathias-lang-sociomantic (approved):
Alright, we can always revisit later.

leandro-lucarella-sociomantic:
@mihails-strasuns-sociomantic fine with you too?

mihails-strasuns-sociomantic (approved):
Yeah, let's not block on nitpicking

PR #66, opened 2017-07-26
Do not output colors for RDMD with DVER=2

llucax:
DMD2 already does it, so by default the path will be in red, and the rest of the message with the right color.


llucax:
Attached code, and of course GitHub made me the author of this issue because of that... 🙄

llucax:
I did some quick testing and it seems to work, but more test is always welcome...

mathias-lang-sociomantic:
I still got a "participating" notification tho.

leandro-lucarella-sociomantic:
Yeah, the author is in reality preserved, you can see it in the issues listing, is just a bug in this view.

Did you try it?

mathias-lang-sociomantic (commented):
> Did you try it?

Just did, works. LGTM.

PR #88, opened 2017-08-02
Allow to specify which directories to convert for d2conv

leandro-lucarella-sociomantic:
Currently, the `d2conv` rule will perform a `find` on `$C`, that is, the top level directory of the git repo in most cases.
We have a use case where we have D2 scripts in the `scripts` folder, which get `d1to2fix`ed anyway, resulting in a diff (and possibly invalid code).

Having a mean to override which directory the `find` will be performed on should be enough to solve this problem.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -131,6 +131,9 @@ HMOD ?= hmod
 # harbored-mod flags
 HMODFLAGS ?= --project-name $(PROJECT_NAME) --project-version $(VERSION)
 
+# Default d1to2fix binary location
+D1TO2FIX ?= d1to2fix
+
 # Default fpm binary location
 FPM ?= fpm
 
@@ -266,6 +269,9 @@ PKGVERSION := $(shell echo $(VERSION) | \
 # For more information refer to the documentation:
 # http://www.gnu.org/software/make/manual/make.html#Text-Functions
 TEST_FILTER_OUT := $C/$(SRC)/%/main.d
+
+# Directories to search for files to convert using d1to2fix
+D1TO2FIX_DIRS := $C
 
 
 # Functions
@@ -686,12 +692,12 @@ fasttest: $(fasttest)
 .PHONY: d2conv
 d2conv: $O/d2conv.stamp
 
-$O/d2conv.stamp: $C
-	$Vfind $C -type f -regex '^.+\.d$$' > $@
-ifeq "$(shell d1to2fix --help 2>/dev/null | grep -- --input)" ""
-	$(call exec, d1to2fix --fatal `cat $@`)
+$O/d2conv.stamp: $(D1TO2FIX_DIRS)
+	$Vfind $(D1TO2FIX_DIRS) -type f -regex '^.+\.d$$' > $@
+ifeq "$(shell $(D1TO2FIX) --help 2>/dev/null | grep -- --input)" ""
+	$(call exec, $(D1TO2FIX) --fatal `cat $@`)
 else
-	$(call exec, d1to2fix $(DIMPORTPATHS) --fatal --input=$@)
+	$(call exec, $(D1TO2FIX) $(DIMPORTPATHS) --fatal --input=$@)
 endif
 
 # Automatic dependency handling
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
Code attached (and issue stolen as usual thanks to the GitHub bug).

Makd.mak
```diff
@@ -267,6 +270,9 @@ PKGVERSION := $(shell echo $(VERSION) | \
 # http://www.gnu.org/software/make/manual/make.html#Text-Functions
 TEST_FILTER_OUT := $C/$(SRC)/%/main.d
 
+# Directories to search for files to convert using d1to2fix
+D1TO2FIX_DIRS := $C
```
mathias-lang-sociomantic, line 274: Should this be `:=` or `?=` ? Since if you are using this variable, there is no way you want to add to it at the moment (since everything is being converted by default).

mathias-lang-sociomantic (commented):
A small comment on the default, otherwise tested and LGTM.

Makd.mak
leandro-lucarella-sociomantic: There is a bit of inconsistency in the use of `:=` vs `?=` I was just following the variables close by. `:=` is a bit misleading though, because if you pass a variable via command-line for example, it will override variables defined with `:=` anyway. `?=` is only for variables defined inside the makefile and since the `Build.mak` is processed after this was defined, you can still override it.

It won't be definable via `Config.mak`, which is where I think the things got a bit more weird. But since this is a pre-existing problem, I just want to be consistent with the other similar variables (like the `TEST_FILTER_OUT` above) for now, to avoid further confusion.

At some point we should review how variables are defined and try to be more consistent and conscious about what should be defined in `Config.mak` and what in `Build.mak`. Is a hard job with little gain, though, and I guess that's why it was never addressed.

leandro-lucarella-sociomantic:
There is one more issue that I noted though, and it's about the include files. We are calling `d1to2fix $(DIMPORTPATHS) --fatal --input=$@`. I guess those import paths should be D1 only, or `d1to2fix` doesn't really care? Do you know?

/cc @mihails-strasuns-sociomantic

Makd.mak
mathias-lang-sociomantic: > It won't be definable via Config.mak, which is where I think the things got a bit more weird. 

Not sure I follow you here ?
Config.mak is included at line 47, so it seems to me that if I do `D1TOFIX_DIRS ?= blah blah2` in my `Config.mak`, this will be used by default unless something is provided on the command line ? That's what we do for `dmd-transitional` in applications ATM.

leandro-lucarella-sociomantic: I mean with `:=`. Again, what I'm saying here is we already have other variables that are supposed to be overrideables defined with `:=`.

The dependencies here are complicated, because there are some things that are supposed to be overrideable via Config.mak and other that are not, because they depedend on the things defined in Config.mak to calculate them. Again, I'm not saying using `?=` won't work, I'm just saying that this is a pre-existing problem that to be solved it will need a careful review of all the variables and where they should be overridden, something that I don't want to address in this PR, and why I used the same operator being used in this section.

leandro-lucarella-sociomantic:
Can you give this a try BTW?

mihails-strasuns-sociomantic:
Import paths must be exactly the same as import paths that would be used to build the binary with D2 compiler. They are used by `libdparse` to resolve symbol names correctly.

leandro-lucarella-sociomantic:
OK, so this should magically work then I guess...

leandro-lucarella-sociomantic:
BTW, @mathias-baumann-sociomantic and @mihails-strasuns-sociomantic it would be nice if you approve the PR if you see it's fine (and in particular if you tested it).

mathias-lang-sociomantic:
> BTW, @mathias-baumann-sociomantic and @mihails-strasuns-sociomantic it would be nice if you approve the PR if you see it's fine (and in particular if you tested it).

Can I do it too ? :trollface:

mathias-lang-sociomantic (commented):
LGTM

leandro-lucarella-sociomantic:
Any mathiases would do :P

PR #90, opened 2017-08-23
Add targets to build examples

### LINKED ISSUE #59: Introduce standard `example` folder

leandro-lucarella-sociomantic:
The new 'example' target will build all examples found in `example/` (by
default). Also the `test` target will build them. The new 'example-run'
target is used to run the examples, but this is not done automatically
by the `test` target as it could be problematic.

Fixes #59.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -59,6 +59,9 @@ __dummy_integrationtest_warning := $(shell echo "MakD Warning: The default \
 endif
 INTEGRATIONTEST ?= test
 
+# Default location to look for examples
+EXAMPLE ?= example
+
 # Check flavours
 FLAVOR_IS_VALID_ := $(if $(filter $F,$(VALID_FLAVORS)),1,0)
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
Anyone? Last pending issue/PR for v1.10.0 / v2.0.0

mihails-strasuns-sociomantic (approved):
Tested with turtle examples, works as expected. Feature-wise can't find anything lacking either for my purposes at least.

PR #94, opened 2017-09-04
Repeated `find` error messages for builds of projects with no integration tests

leandro-lucarella-sociomantic:
When attempting to build a project without integration tests using v2.0.0-rc1, the following error message occurs multiple times in the build log:

```
find: ‘/home/me/projectdir/integrationtest’: No such file or directory
```

This does not happen for v1.x.x, and feels a bit intrusive, so it would be nice if `find` could fail silently in these cases.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -273,6 +273,12 @@ D1TO2FIX_DIRS := $C
 # returns empty.
 eq = $(if $(subst $1,,$2),,$1)
 
+# Simple wrapper to call find, only if the directory(ies) where to look exist
+# $1 find directories
+# $2 find options
+find = $(foreach d,$1,$(if $(shell test -d "$d" && echo true),\
+	   $(shell find $1 $2)))
+
 # Find files and get the their file names relative to another directory.
 # $1 is the files suffix (".h" or ".cpp" for example).
 # $2 is a directory rewrite, the matched files will be rewriten to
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
I will move this to v1.9.x because is a more general problem.

leandro-lucarella-sociomantic:
Code attached.

leandro-lucarella-sociomantic:
Some testing might be needed. It's a bug fix but not a minor one. Maybe it should be included in v2 anyway, as it's not very important...

joseph-wakeling-sociomantic:
> I will move this to v1.9.x because is a more general problem.

That's weird, I never noticed it before trying out v2.0.0-rc1.

leandro-lucarella-sociomantic:
There are many instances where this could happen. Maybe before it couldn't happen for the integration tests though.

leandro-lucarella-sociomantic:
But this is another reason to move to v2.0.0. I will do that.

leandro-lucarella-sociomantic:
Moved back to v2. Please review so we can release v2.0.0.

@joseph-wakeling-sociomantic if you could test your projects with this it would be good so we don't release a buggy v2 with these last minute changes.

mihails-strasuns-sociomantic (approved):
LGTM, waiting for @joseph-wakeling-sociomantic confirmation

joseph-wakeling-sociomantic:
Seems to work fine from my side :-)

PR #98, opened 2017-09-12
List build flags / flavour in version info

leandro-lucarella-sociomantic:
It would be really useful, say, to see if a binary was built with `-release` and so on.

### DIFF
--- a/Version.d.tpl
+++ b/Version.d.tpl
@@ -46,6 +46,8 @@ static this()
     version_info["build_date"] = "@DATE@";
     version_info["build_author"] = "@AUTHOR@";
     version_info["compiler"] = "@COMPILER@";
+    version_info["dflags"] = "@DFLAGS@";
+    version_info["flavour"] = "@FLAVOUR@";
 @LIBRARIES@
     buildPredefinedVersions();
 }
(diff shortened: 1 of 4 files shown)


leandro-lucarella-sociomantic:
Fix attached.

Version.d.tpl
```diff
@@ -46,6 +46,8 @@ static this()
     version_info["build_date"] = "@DATE@";
     version_info["build_author"] = "@AUTHOR@";
     version_info["compiler"] = "@COMPILER@";
+    version_info["dflags"] = "@DFLAGS@";
```
mihails-strasuns-sociomantic, line 49: This may be confusing as it won't include flags supplied to DMD indirectly via `RDMDFLAGS` or some other env var. Is that OK?

gavin-norman-sociomantic:
Thanks! I guess we'll need to hook this into the app infrastructure in ocean, in order to print this stuff when the app is run with `--version`?

Version.d.tpl
leandro-lucarella-sociomantic: Yeah, I think including `RDMDFLAGS` is more noise than anything else, because it's all about the build system, not how code is actually compiled (unless `RDMDFLAGS` are misused to pass `DFLAGS`). Currently `RDMDFLAGS` passed automatically are `--compiler` (already present in `Version`) and `--force` (irrelevant).

leandro-lucarella-sociomantic:
> Thanks! I guess we'll need to hook this into the app infrastructure in ocean, in order to print this stuff when the app is run with --version?

Oh, right, since the `version_info` is just an associative array I thought everything there was printed automatically. :-/

leandro-lucarella-sociomantic:
https://github.com/sociomantic-tsunami/ocean/pull/247

mihails-strasuns-sociomantic (approved):

PR #99, opened 2017-09-14
Add coverage support

leandro-lucarella-sociomantic:
When using `make COV=1` the code will be compiled with `-cov`, and will
generate coverage reports when running.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -279,6 +279,22 @@ TEST_FILTER_OUT := $C/$(SRC)/%/main.d
 D1TO2FIX_DIRS := $C
 
 
+# Code coverage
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
Even when this is already useful, it should also wait until https://github.com/sociomantic-tsunami/tangort/pull/21 and https://github.com/dlang/druntime/pull/1916 are merged, just to make sure the env variable names are right.

leandro-lucarella-sociomantic:
Should be moved to v2.x.x

leandro-lucarella-sociomantic:
Duplicate of #115

PR #100, opened 2017-09-18
Add branch name to version string if possible

gautam-kotian-sociomantic:
This was always meant to work, but never did reliably in practice. This should hopefully improve things without causing any differences to the version string when dealing with tagged commits.


gautam-kotian-sociomantic:
Sorry, messed up with the user name. Will recreate.

PR #101, opened 2017-09-18
Make version string more concise when possible

gkotian:
This was always meant to work, but never did reliably in practice. This should hopefully improve things without causing any differences to the version string when dealing with tagged commits.

### DIFF
--- a/mkversion.sh
+++ b/mkversion.sh
@@ -58,9 +58,9 @@ do
 done
 shift `expr $OPTIND - 1`
 
-version=`git describe --dirty --always`
+version=`git describe --dirty --always --abbrev=4`
 # Add branch name if we only got a hash
-echo "$version" | grep -Eq '^[0-9a-f]{7}(-dirty)?$' &&
+echo "$version" | grep -Eq '^[0-9a-f]{4,}(-dirty)?$' &&
     version=`git rev-parse --abbrev-ref HEAD`-g"$version"
 
 # Check if the user only wanted to print the version number
@@ -96,9 +96,9 @@ for lib in "$@"
 do
     lib_base=`basename $lib`
 
-    ver_desc=`cd $lib && git describe --dirty --always`
+    ver_desc=`cd $lib && git describe --dirty --always --abbrev=4`
     # Add branch name if we only got a hash
-    echo "$ver_desc" | grep -Eq '^[0-9a-f]{7}(-dirty)?$' &&
+    echo "$ver_desc" | grep -Eq '^[0-9a-f]{4,}(-dirty)?$' &&
         ver_desc=`cd $lib && git rev-parse --abbrev-ref HEAD`-g"$ver_desc"
 
     libs="${libs}    versionInfo[\"lib_${lib_base}\"] = \"${ver_desc}\";\\n"
--- a/mkversion.sh
+++ b/mkversion.sh
@@ -60,7 +60,7 @@ shift `expr $OPTIND - 1`
 
 version=`git describe --dirty --always`
 # Add branch name if we only got a hash
-echo "$version" | egrep -q '^[0-9a-f]{7}(-dirty)?$'  &&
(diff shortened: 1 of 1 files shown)


mkversion.sh
```diff
@@ -60,7 +60,7 @@ shift `expr $OPTIND - 1`
 
 version=`git describe --dirty --always --abbrev=7`
 # Add branch name if we only got a hash
-echo "$version" | egrep -q '^[0-9a-f]{7}(-dirty)?$'  &&
+echo "$version" | egrep -q '^.*[0-9a-f]{7}(-dirty)?$'  &&
```
leandro-lucarella-sociomantic: This fixed 7 chars is actually wrong, not the `--abbrev` in the previous commit, as the man of `git describe` says:

>       --abbrev=<n>
>           Instead of using the default 7 hexadecimal digits as the abbreviated object name, use <n> digits, or as many digits as needed
>           to form a unique object name. An <n> of 0 will suppress long format, only showing the closest tag.

So it could happen that is actually more. I guess using an explicit `--abbrev` is a good thing too, but this regexp should probably be then `{7,}` (we could make it `{7,40}` but maybe git hashes will grow in the future and I'm not sure if it provides any advantages to have an upper limit).

We could make the explicit `--abbrev` use `4` too, I guess the more concise the version the better.

leandro-lucarella-sociomantic: BTW, probably the bug here is to start the regex with `^`, as we only want to match the end of the string. :)

leandro-lucarella-sociomantic: Wait! This is supposed to detect when only a hash was given. It shouldn't match a branch name too. If there is a bug, it is elsewhere.

leandro-lucarella-sociomantic:
I am doing some tests and `mkversion` works as expected (without applying this PR) for me in the tests I'm doing (a repo with no tags, where `git describe --dirty --always --abbrev=7` gives you just a hash).

Can you say in which scenario this is failing for you?

gkotian:
I also get only a hash on a repo without tags. But when trying it on a repo with tags, I get something like `v1.0.0-18-gc12687c` if I'm on a commit that is 18 commits after the `v1.0.0` tag.

mkversion.sh
gkotian: Hmm... so do you suggest changing the regex to contain `{7,}` and removing `^.*`? So this line will become:
```
echo "$version" | egrep -q '[0-9a-f]{7,}(-dirty)?$' &&
```
I haven't tested this, but it seems reasonable.

leandro-lucarella-sociomantic:
> I also get only a hash on a repo without tags.

In this case, you get just a hash with `git describe --dirty --always --abbrev=7`, but when doing `git describe --dirty --always --abbrev=7 | egrep -q '^[0-9a-f]{7,}(-dirty)?$' && git rev-parse --abbrev-ref HEAD` don't you get the branch name?

leandro-lucarella-sociomantic:
If your hash wasn't unique with only 7 chars, maybe you are getting a hash that is more than 7 chars and because of that, and only that, `egrep -q '^[0-9a-f]{7}(-dirty)?$'`, maybe that's it?

mkversion.sh
leandro-lucarella-sociomantic: I did before I found out that the regex should check for the whole line. Because is actually checking if `git describe` is giving **just a hash**, so it should be kept as `^[0-9a-f]{7,}(-dirty)?$` (the extra `,` is needed because the hash could be actually be longer if 7 chars are not enough to disambiguate the commit).

gkotian:
I have changed this PR to not deal with the branch name any more. Now it only adds the `--abbrev=4` argument to `git describe` in order to try and make the version more concise.

Changed base to v1.x.x as this is no longer a bug fix.

leandro-lucarella-sociomantic (approved):
Perfect, thanks! Can you rebase it on top of v1.9.x?

gautam-kotian-sociomantic:
Rebased on top of v1.9.x

PR #102, opened 2017-10-11
add fmt target

stefan-koch-sociomantic:
adding format target `fmt` to makd
currently this does format with `upstream`/`phobos` style.

as getting closer to sociomantic style requires a change to dfmt.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -123,6 +123,12 @@ HMOD ?= hmod
 # harbored-mod flags
 HMODFLAGS ?= --project-name $(PROJECT_NAME) --project-version $(VERSION)
 
+# Default code-formatter (dfmt)
+DFMT ?= dfmt
+
+# dfmt flags
+DFMTFLAGS ?= --space_before_function_parameters=true
+
 # Default d1to2fix binary location
 D1TO2FIX ?= d1to2fix
 
@@ -661,6 +667,17 @@ $O/doc.stamp: $(call find,$(SRC),-type f \( -name '*.d' -o -name '*.di' \))
 	$Vtouch $@
 
 doc += $O/doc.stamp
+
+# Formatter rules
+###################
+
+# General rule to run the dfmt formatter 
+$O/dfmt.stamp: $(shell mkdir -p formatted)
+	$(shell find $(SRC) -type f \( -name '*.d' -o -name '*.di' \))
+	$(call exec,$(DFMT) $(DFMTFLAGS) $(SRC) > formatted/$(SRC) ,fmt)
+	$Vtouch $@
+
+dfmt += $O/dfmt.stamp
 
 # Graph dependencies rule
 ##########################


Makd.mak
```diff
@@ -662,6 +662,13 @@ $O/doc.stamp: $(call find,$(SRC),-type f \( -name '*.d' -o -name '*.di' \))
 
 doc += $O/doc.stamp
 
+# General rule to run the harbored-mod generator
+$O/dfmt.stamp: $(shell find $(SRC) -type f \( -name '*.d' -o -name '*.di' \))
+	$(call exec,dfmt -i $(SRC) >/dev/null,fmt)
+	$Vtouch $@
+
+dfmt += $O/dfmt.stamp
```
leandro-lucarella-sociomantic: This should probably be `dfmt: $O/dfmt.stamp`. But I suggest using `lint: $O/dfmt.stamp`, in case we can add more style checkers / linters in the future...

Makd.mak
```diff
@@ -662,6 +662,13 @@ $O/doc.stamp: $(call find,$(SRC),-type f \( -name '*.d' -o -name '*.di' \))
 
 doc += $O/doc.stamp
 
+# General rule to run the harbored-mod generator
+$O/dfmt.stamp: $(shell find $(SRC) -type f \( -name '*.d' -o -name '*.di' \))
```
leandro-lucarella-sociomantic: Please use `$C/$(SRC)`.

Makd.mak
```diff
@@ -662,6 +662,13 @@ $O/doc.stamp: $(call find,$(SRC),-type f \( -name '*.d' -o -name '*.di' \))
...
+	$(call exec,dfmt -i $(SRC) >/dev/null,fmt)
```
leandro-lucarella-sociomantic: From experience, it's better to have an easy way to override the `dfmt` binary to use, add a variable `DFMT ?= dfmt` as we have for `D1TO2FIX`, etc.

Also, isn't it possible to pass the files explicitly so there are no inconsistency between the dependencies and the files the command runs on? If so, please use `$(call exec,$(DFMT) -i $^ >/dev/null,dfmt,$(SRC))` or something like that.

leandro-lucarella-sociomantic (commented):
So, this will basically change the files. For CI for me this is fine, but to run on users workstations, I think it is quite disruptive to get files changed. It would be better to have some way to write files to a temporary location and run a diff to see if there are changes (or have a `link` target that does that and a `fixstyle` target that actually changes the code).

Also if we have a target that is non-destructive and fails if the lint fails, we can promote the usage of a git pre-commit hook so code can't even be committed if the style is wrong.

stefan-koch-sociomantic:
@leandro-lucarella-sociomantic thanks for the review.
I will address the comments. (unfortunately my makefile-mojo is quite weak)

Makd.mak
```diff
@@ -662,6 +668,17 @@ $O/doc.stamp: $(call find,$(SRC),-type f \( -name '*.d' -o -name '*.di' \))
 
 doc += $O/doc.stamp
 
+# Formatter rules
+###################
+
+# General rule to run the dfmt formatter 
+$O/dfmt.stamp: $(shell mkdir -p formatted)
```
leandro-lucarella-sociomantic, line 675: Mmmm, this doens't make sense (the `$(shell ...)` part as a dependency).

Makd.mak
```diff
@@ -662,6 +668,17 @@ $O/doc.stamp: $(call find,$(SRC),-type f \( -name '*.d' -o -name '*.di' \))
...
+	$(shell find $(SRC) -type f \( -name '*.d' -o -name '*.di' \))
```
leandro-lucarella-sociomantic, line 676: This doesn't make sense either. Maybe you intended to have this shell line and the one above swapped?

Makd.mak
```diff
@@ -662,6 +668,17 @@ $O/doc.stamp: $(call find,$(SRC),-type f \( -name '*.d' -o -name '*.di' \))
...
+	$(call exec,$(DFMT) $(DFMTFLAGS) $(SRC) > formatted/$(SRC) ,fmt)
```
leandro-lucarella-sociomantic, line 677: This doesn't make sense either :)

`$(SRC)` is a directory, so you can't redirect files to a directory.

Depending on what options `dfmt` supports, you might want to convert one file at the time instead (it might be slower though):

```make
$O/formatted/%: $(SRC)/%
    $(call exec,$(DFMT) $(DFMTFLAGS) $^ > $@,fmt)
```

This is a *pattern rule* and tell make how to build a file based on another file and a pattern. You still need to create a rule to list all the target files. But then I wonder what is the purpose of storing all these files? Is to make a diff?

Makd.mak
```diff
@@ -662,6 +668,17 @@ $O/doc.stamp: $(call find,$(SRC),-type f \( -name '*.d' -o -name '*.di' \))
...
+	$Vtouch $@
+
+dfmt += $O/dfmt.stamp
```
leandro-lucarella-sociomantic, line 680: This doesn't do much, you probably wanted to define a rule instead:

```make
dfmt : $O/dfmt.stamp
```

leandro-lucarella-sociomantic (changes requested):
Also, the commit title should be Capitalized :)

joseph-wakeling-sociomantic:
Target milestone seems wrong for this PR?

leandro-lucarella-sociomantic:
Assigned to *Future* until this gains traction again.

leandro-lucarella-sociomantic:
We don't plan to use `dfmt` anymore, so closing.

Issue #104, opened 2017-10-16
The packages to build should be overrideable


mathias-lang-sociomantic:
Currently it's not possible to filter out some package to have them not being built, because the `pkg` rule is: https://github.com/sociomantic-tsunami/makd/blob/602ca95549d6290e31fa96d1a119a452089c5d32/Makd.mak#L490

It would be nice to move this to a variable.

PR #105, opened 2017-10-18
 Allow to exclude some packages from building

mathias-lang-sociomantic:
> Some users might have packages that can only be built in a certain configuration, e.g. a certain flavor, D1/D2 only...
> In order for account for those cases, one can filter-out the PKGS variable (or override it).

Also defines support term for major branch since we now have 2.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -151,6 +151,9 @@ PKGTYPE ?= deb
 # Default flags to pass to graph-deps (see --help for details)
 GRAPH_DEPS_FLAGS ?= -c -C
 
+# Packages files to build. Each file builds a different package.
+PKG_FILES ?= $(wildcard $(PKG)/*.pkg)
+
 # Default fpm flags. By default it builds Debian packages from files, a Debian
 # changelog is provided, and a version and iteration (using the Debian version)
 # (defined lazily so we can use target variables)
@@ -485,9 +488,6 @@ $B/%: $G/build-d-flags
 clean:
 	$(call exec,$(RM) -r $(VD) $(clean),$(VD) $(clean))
 
-# Phony target to build all packages
-.PHONY: pkg
-pkg: $(patsubst $(PKG)/%.pkg,$O/pkg-%.stamp,$(wildcard $(PKG)/*.pkg))
 
 # Target to build a package based on the fpm definition (old packages are
 # removed to avoid infinite pollution of the build directory, as every package
@@ -744,6 +744,11 @@ test: $(test)
 # to build and run tests to the $(fasttest) variable).
 .PHONY: fasttest
 fasttest: $(fasttest)
+
+# Phony rule to build all packages
+pkg += $(patsubst $(PKG)/%.pkg,$O/pkg-%.stamp,$(PKG_FILES))
+.PHONY: pkg
+pkg: $(pkg)
 
 
 # Temporary rule to convert code from D1 to D2
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,7 @@ for versioning.
 Support Guarantees
 ------------------
 
-* Major branch development period: undefined
+* Major branch development period: 3 months
 * Maintained minor versions: 2 most recent
 
 Maintained Major Branches
@@ -34,9 +34,12 @@ Maintained Major Branches
 ====== ==================== ===============
 Major  Initial release date Supported until
 ====== ==================== ===============
-v1.x.x v1.4.0_: 24/06/2016  TBD
+v1.x.x v1.4.0_: 24/06/2016  Fri 08/12/2017
+====== ==================== ===============
+v2.x.x v2.0.0_: 08/09/2017  TBD
 ====== ==================== ===============
 .. _v1.4.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v1.4.0
+.. _v2.0.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v2.0.0
 
 
 .. contents::
@@ -1100,4 +1103,3 @@ be re-compiled in that case!
 
 .. _Makeit: https://git.llucax.com/w/software/makeit.git
 .. _fpm: https://github.com/jordansissel/fpm
-
--- a/README.rst
+++ b/README.rst
@@ -404,6 +404,7 @@ Variables you might want to override
 * ``PKG`` is where package definitions are searched. When building packages,
   each ``*.pkg`` file in that directory will be built. By default ``$T/pkg``.
 * ``PKG_DEFAULTS`` contains the default options passed to ``mkpkg``.
+* ``PKG_FILES`` contains the list of packages definitions.
 * ``PKG_PREBUILD`` hold commands to run previous to build packages.
 * ``PROJECT_NAME`` contains the name of the project, used in documentation
   generatation. It defaults to the name of the top directory.
@@ -683,9 +684,13 @@ Packaging
 
 Makd supports a simple facility to make packages based on fpm_.  A simple
 wrapper program ``mkpkg`` is provided to ease the creation of scripts that use
-fpm_ to create packages.  The predefined ``pkg`` target will scan for ``*.pkg``
-files in the ``$(PKG)`` directory (by default ``$T/pkg``) and then invoke
-``mkpkg`` with them.
+fpm_ to create packages.
+The predefined ``pkg`` target depends on the special variable ``$(pkg)``,
+where you can add any extra target that must be built for ``pkg``.
+By default every package file specified in `$(PKG_FILES)` will be
+added to ``$(pkg)``, and by default ``$(PKG_FILES)`` holds all
+``*.pkg`` files in the ``$(PKG)`` directory (by default ``$T/pkg``).
+``mkpkg`` is invoked for every file specified in ``$(PKG_FILES)``.
 
 These files are expected to be Python scripts defining two variables:
 
@@ -768,6 +773,15 @@ variable ``FUN``:
 
         Note that ``OPTS['description']`` must be defined and hold a non-empty
         string.
+
+One can exclude packages from being built under certain conditions
+(e.g. D1/D2 only packages) by filtering ``PKG_FILES``:
+
+.. code:: Makefile
+
+          ifeq ($(DVER),1)
+          PKG_FILES := $(filter-out $(PKG)/D2OnlyApp.pkg,$(PKG_FILES))
+          endif
 
 Generated packages will be stored in the ``$P`` directory (by default
 ``$G/pkg``. Since each package usually have a different name, as the version
--- a/relnotes/pkgs-variable.feature.md
+++ b/relnotes/pkgs-variable.feature.md
@@ -1,0 +1,11 @@
+* `PKG_FILES`
+
+  This variable holds all package definitions to be build. The purpose is to allow one to exclude
+  some packages from being built if need be, for example for `F=production` or `DVER=2`
+  only packages.
+  It can be filtered-out, e.g:
+  ```
+  ifeq ($(DVER),1)
+  PKG_FILES := $(filter-out $(PKG)/D2OnlyApp.pkg,$(PKG_FILES))
+  endif
+  ```


README.rst
```diff
@@ -25,7 +25,7 @@ for versioning.
 Support Guarantees
 ------------------
 
-* Major branch development period: undefined
+* Major branch development period: 6 months
```
leandro-lucarella-sociomantic: And this is based on what? :)

leandro-lucarella-sociomantic: And this is based on what? :)
I think 6 months could be a bit stretched for Makd, is just a build thing, it should pose very little risk to upgrade, so I would go for something more in the lines of 3 months.

Makd.mak
```diff
@@ -487,7 +487,8 @@ clean:
 
 # Phony target to build all packages
 .PHONY: pkg
-pkg: $(patsubst $(PKG)/%.pkg,$O/pkg-%.stamp,$(wildcard $(PKG)/*.pkg))
+PKGS ?= $(wildcard $(PKG)/*.pkg)
```
leandro-lucarella-sociomantic: All variables used to accumulate dependencies of a target are in lowercase so far, and maked after the target (`all`, `test`, etc.), I would use simply `pkg`.

I wonder if it should be `+=` instead, as it seems like the most common case will be to want to `$(filter-out )` something instead of building the list from the scratch.

Makd.mak
```diff
@@ -487,7 +487,8 @@ clean:
...
+pkg: $(patsubst $(PKG)/%.pkg,$O/pkg-%.stamp,$(PKGS))
```
leandro-lucarella-sociomantic: Mmm, I see that you are still leaving the `$(patsubst )` in the rule. I think then it might be better to do the following to have both convenience and full flexibility:

* Use a `PKG_FILES` variable to simply specify package files to build, but define it *before* `Build.mak` is included (`PKG_FILES := $(wildcard $(PKG)/*.pkg)` so it can be overridden and simply `$(filter-out )`ed).
* Use a `pkg` variable to accumulate dependencies for the `pkg` target (like real targets)
* Assign the converted `PKG_FILES` as `pkg` targets `pkg += $(patsubst $(PKG)/%.pkg,$O/pkg-%.stamp,$(PKG_FILES))`

This way you can define the `PKG_FILES` from scratch, or you can `$(filter-out )` the defaults, and yet you can have any other crazy method to build packages and add/remove dependencies to `make pkg`.

leandro-lucarella-sociomantic (changes requested):

README.rst
mathias-lang-sociomantic: I just went with what ocean (and IIRC other core team projects) is doing. I'm fine with 3 months and will update to that.

mathias-lang-sociomantic:
Thanks for the review @leandro-lucarella-sociomantic . It looks so much better now.

README.rst
```diff
@@ -34,9 +34,12 @@ Maintained Major Branches
 ====== ==================== ===============
 Major  Initial release date Supported until
 ====== ==================== ===============
-v1.x.x v1.4.0_: 24/06/2016  TBD
+v1.x.x v1.4.0_: 24/06/2016  Fri 08/12/2017
+====== ==================== ===============
+v2.x.x v2.0.1_: 11/09/2017  TBD
 ====== ==================== ===============
 .. _v1.4.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v1.4.0
+.. _v2.0.1: https://github.com/sociomantic-tsunami/makd/releases/tag/v2.0.1
```
leandro-lucarella-sociomantic: There is an inconsistency in here, the end of support for v1 is 08/12/2017 but you are putting the release of v2 on 11/09/2017. I think you got the date from the release of v2.0.0, not v2.0.1, and, even when broken, I think the real initial release should be put here, i.e. v2.0.0.

Makd.mak
```diff
@@ -151,6 +151,9 @@ PKGTYPE ?= deb
 # Default flags to pass to graph-deps (see --help for details)
 GRAPH_DEPS_FLAGS ?= -c -C
 
+# Packages files to build. Each file builds a different package.
+PKG_FILES := $(wildcard $(PKG)/*.pkg)
```
leandro-lucarella-sociomantic: Maybe `?=` to match the other variables around?

Makd.mak
```diff
@@ -745,6 +745,11 @@ test: $(test)
 .PHONY: fasttest
 fasttest: $(fasttest)
 
+# Phony rule to build all packages
+.PHONY: pkg
+pkg += $(patsubst $(PKG)/%.pkg,$O/pkg-%.stamp,$(PKG_FILES))
+pkg: $(pkg)
+
```
leandro-lucarella-sociomantic: Just for clarity, I would put the variable definition before the `.PHONY: pkg`, this is the style used across the file (group the target definition with the declaration of being a phony target).

leandro-lucarella-sociomantic (commented):
Looks much better indeed. Can you also add this info to the README please?

README.rst
mathias-lang-sociomantic: Actually I wanted to align it on a week's boundary. But yeah, this is a weird mix, will fix.

mathias-lang-sociomantic:
Updated according to @leandro-lucarella-sociomantic 's comments

README.rst
```diff
@@ -34,9 +34,12 @@ Maintained Major Branches
 ====== ==================== ===============
 Major  Initial release date Supported until
 ====== ==================== ===============
-v1.x.x v1.4.0_: 24/06/2016  TBD
+v1.x.x v1.4.0_: 24/06/2016  Fri 08/12/2017
+====== ==================== ===============
+v2.x.x v2.0.0_: 08/09/2017  TBD
 ====== ==================== ===============
 .. _v1.4.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v1.4.0
+.. _v2.0.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v2.0.1
```
leandro-lucarella-sociomantic: You missed the link :smiling_imp:

leandro-lucarella-sociomantic (changes requested):
Also still missing the updates to the README... :police_car:

mathias-lang-sociomantic:
Would fix #104 BTW.

Updated:
```diff
diff --git a/README.rst b/README.rst
index 680d179..db1e209 100644
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ v1.x.x v1.4.0_: 24/06/2016  Fri 08/12/2017
 v2.x.x v2.0.0_: 08/09/2017  TBD
 ====== ==================== ===============
 .. _v1.4.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v1.4.0
-.. _v2.0.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v2.0.1
+.. _v2.0.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v2.0.0
 
 
 .. contents::
 1 file changed, 1 insertion(+), 1 deletion(-)
geod24@labs-164 ~/tsunami/makd (git)-[stfu] % ne README.rst 
geod24@labs-164 ~/tsunami/makd (git)-[stfu] % git diff
diff --git a/README.rst b/README.rst
index db1e209..379f002 100644
--- a/README.rst
+++ b/README.rst
@@ -404,6 +404,7 @@ Variables you might want to override
 * ``PKG`` is where package definitions are searched. When building packages,
   each ``*.pkg`` file in that directory will be built. By default ``$T/pkg``.
 * ``PKG_DEFAULTS`` contains the default options passed to ``mkpkg``.
+* ``PKG_FILES`` contains the list of packages definitions.
 * ``PKG_PREBUILD`` hold commands to run previous to build packages.
 * ``PROJECT_NAME`` contains the name of the project, used in documentation
   generatation. It defaults to the name of the top directory.
```

mathias-lang-sociomantic:
Ping @leandro-lucarella-sociomantic

README.rst
```diff
@@ -404,6 +404,7 @@ Variables you might want to override
 * ``PKG`` is where package definitions are searched. When building packages,
   each ``*.pkg`` file in that directory will be built. By default ``$T/pkg``.
 * ``PKG_DEFAULTS`` contains the default options passed to ``mkpkg``.
+* ``PKG_FILES`` contains the list of packages definitions.
```
leandro-lucarella-sociomantic, line 407: I would add a bit more than this in the Packaging section. Like the example in the release notes. Also I would mention the `pkg` variable as a more general approach to add stuff to do for the `make pkg` target.

leandro-lucarella-sociomantic (changes requested):
I think the README needs a bit further improvement and is good to go.

mathias-lang-sociomantic:
Updated:
```diff
diff --git a/README.rst b/README.rst
index 379f002..174df7d 100644
--- a/README.rst
+++ b/README.rst
@@ -770,6 +770,19 @@ variable ``FUN``:
         Note that ``OPTS['description']`` must be defined and hold a non-empty
         string.
 
+One package is generated by package definition file. The package definition
+file used by default will be in ``PKG_FILES``. This is useful if one wants
+to exclude certain packages from being built, e.g. for D1 or D2-only packages:
+
+.. code:: Makefile
+
+          ifeq ($(DVER),1)
+          PKG_FILES := $(filter-out $(PKG)/D2OnlyApp.pkg,$(PKG_FILES))
+          endif
+
+Beyond the ``PKG_FILES`` variable, the ``pkg`` variable can be used as a
+more general approach to add dependencies for ``make pkg``.
+
 Generated packages will be stored in the ``$P`` directory (by default
 ``$G/pkg``. Since each package usually have a different name, as the version
```

README.rst
```diff
@@ -766,6 +770,19 @@ variable ``FUN``:
         Note that ``OPTS['description']`` must be defined and hold a non-empty
         string.
 
+One package is generated by package definition file. The package definition
+file used by default will be in ``PKG_FILES``. This is useful if one wants
```
leandro-lucarella-sociomantic: To be honest the first 2 sentences are not clear to me. And also there is a part above where it says all packages in `pkg/*` will be built:
```rst
The predefined ``pkg`` target will scan for ``*.pkg`` files in the ``$(PKG)``
directory (by default ``$T/pkg``) and then invoke ``mkpkg`` with them.
```

What about moving this part next to that and unify if, something like:
```rst
The predefined ``pkg`` target depends on the special variable ``$(pkg)``,
where you can add any extra target that must be built for ``pkg``.
By default every package file specified in `$(PKG_FILES)` will be
added to ``$(pkg)``, and by default `$(PKG_FILES)` holds all
``*.pkg`` files in the ``$(PKG)`` directory (by default ``$T/pkg``).
``mkpkg`` is invoked for every file specified in `$(PKG_FILES)`.
```

Then the example below can be maybe kept here.

mathias-lang-sociomantic:
Updated:

```diff
diff --git a/README.rst b/README.rst
index 0aa17a6..fda5799 100644
--- a/README.rst
+++ b/README.rst
@@ -684,9 +684,13 @@ Packaging
 
 Makd supports a simple facility to make packages based on fpm_.  A simple
 wrapper program ``mkpkg`` is provided to ease the creation of scripts that use
-fpm_ to create packages.  The predefined ``pkg`` target will scan for ``*.pkg``
-files in the ``$(PKG)`` directory (by default ``$T/pkg``) and then invoke
-``mkpkg`` with them.
+fpm_ to create packages.
+The predefined ``pkg`` target depends on the special variable ``$(pkg)``,
+where you can add any extra target that must be built for ``pkg``.
+By default every package file specified in `$(PKG_FILES)` will be
+added to ``$(pkg)``, and by default `$(PKG_FILES)` holds all
+``*.pkg`` files in the ``$(PKG)`` directory (by default ``$T/pkg``).
+``mkpkg`` is invoked for every file specified in `$(PKG_FILES)`.
 
 These files are expected to be Python scripts defining two variables:
 
@@ -770,9 +774,8 @@ variable ``FUN``:
         Note that ``OPTS['description']`` must be defined and hold a non-empty
         string.
 
-One package is generated by package definition file. The package definition
-file used by default will be in ``PKG_FILES``. This is useful if one wants
-to exclude certain packages from being built, e.g. for D1 or D2-only packages:
+One can exclude packages from being built under certain conditions
+(e.g. D1/D2 only packages) by filtering ``PKG_FILES``:
 
 .. code:: Makefile
 
@@ -780,9 +783,6 @@ to exclude certain packages from being built, e.g. for D1 or D2-only packages:
           PKG_FILES := $(filter-out $(PKG)/D2OnlyApp.pkg,$(PKG_FILES))
           endif
 
-Beyond the ``PKG_FILES`` variable, the ``pkg`` variable can be used as a
-more general approach to add dependencies for ``make pkg``.
-
 Generated packages will be stored in the ``$P`` directory (by default
 ``$G/pkg``. Since each package usually have a different name, as the version
 usually changes with each change, all old packages are removed before making
```

README.rst
```diff
@@ -680,9 +684,13 @@ Packaging
...
 Makd supports a simple facility to make packages based on fpm_.  A simple
 wrapper program ``mkpkg`` is provided to ease the creation of scripts that use
-fpm_ to create packages.  The predefined ``pkg`` target will scan for ``*.pkg``
-files in the ``$(PKG)`` directory (by default ``$T/pkg``) and then invoke
-``mkpkg`` with them.
+fpm_ to create packages.
+The predefined ``pkg`` target depends on the special variable ``$(pkg)``,
+where you can add any extra target that must be built for ``pkg``.
+By default every package file specified in `$(PKG_FILES)` will be
+added to ``$(pkg)``, and by default `$(PKG_FILES)` holds all
```
leandro-lucarella-sociomantic: Missing double backtick for both `$(PKG_FILES)` :grimacing:

README.rst
```diff
@@ -680,9 +684,13 @@ Packaging
...
+``*.pkg`` files in the ``$(PKG)`` directory (by default ``$T/pkg``).
+``mkpkg`` is invoked for every file specified in `$(PKG_FILES)`.
```
leandro-lucarella-sociomantic: Here too.

mathias-lang-sociomantic: I copy pasted !!!!11111
Fixed

leandro-lucarella-sociomantic: :scream:

leandro-lucarella-sociomantic (approved):

mathias-lang-sociomantic:
:tada:

PR #110, opened 2017-11-15
Convert release notes to the new format

leandro-lucarella-sociomantic:
Also improve the notes a bit.

### DIFF
--- a/relnotes/build-flags.feature.md
+++ b/relnotes/build-flags.feature.md
@@ -1,3 +1,5 @@
-* Now the generated version module will include the build flags and *flavour*
-  (`$F`) used to compile the project. The new keys are named `dflags` and
-  `flavour` respectively.
+## Compiler flags and *flavour* (`$F`) is now part of `version_info`
+
+`Version`, `mkversion.sh`
+
+Now the generated version module will include the build flags and *flavour* (`$F`) used to compile the project. The new keys are named `dflags` and `flavour` respectively.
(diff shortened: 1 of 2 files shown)


relnotes/build-flags.feature.md
```diff
@@ -1,3 +1,5 @@
-* Now the generated version module will include the build flags and *flavour*
-  (`$F`) used to compile the project. The new keys are named `dflags` and
-  `flavour` respectively.
+## Compiler flags and *flavour* (`$F`) is now part of `version_info`
+
+`version_info`, `mkversion.sh`
```
mathias-lang-sociomantic: `Version` ? The guidelines say "modules" affected.

relnotes/pkgs-variable.feature.md
```diff
@@ -1,11 +1,15 @@
...
+## Packages to be build can be specified via `PKG_FILES`
+
+`mkpkg`
+
+A new `PKG_FILES` variable holds all package definitions to be build.
+
+The purpose is to allow one to exclude some packages from being built if need be, for example for `F=production` or `DVER=2` only packages.
+
+The `PKG_FILES` variable comes populated with `pkg/*.pkg` by default, but it can
+be also filtered-out, e.g:
```
mathias-lang-sociomantic: The "also" makes little sense to me. Maybe you mean "[...] by default, it can be overriden or filtered out as follows:", or "[...] by default, and can be filtered out as follows:"

leandro-lucarella-sociomantic: No, I mean it like that. I think I already had a discussion with a similar case. I guess this is probably a valid use if translated literally to Spanish, but something that doesn't work out in English :P

I will use something similar to what you suggest.

leandro-lucarella-sociomantic:
Updated

mathias-lang-sociomantic (approved):
LGTM

Issue #111, opened 2017-11-23
Makd v1.11.0 / v2.1.0 are broken


mathias-lang-sociomantic:
Running MakD v2.1.0 will yield `sed: -e expression #6, char 29: unknown option to `s'`.
Moreover, `Version.d`'s module name is not replaced, leading to the following code:
```
module @MODULE@;
```

Which obviously fails to compile.

leandro-lucarella-sociomantic:
This potentially affects v1.10.x and v2.0.x too (if some funny chars come in them).

mathias-lang-sociomantic:
Tested on my side, v2.1.1 works fine

leandro-lucarella-sociomantic:
Releasing! Sorry for the trouble!

PR #112, opened 2017-11-23
Fix for replacing `@DFLAGS@` with build flags

jens-mueller-sociomantic:
This change replaces the sed delimiter used for replacing the string
`@DFLAGS@` with the contents of `$DFLAGS` to use `\#` instead of `/`.
This is safer because `#` is unlikely to be in `$DFLAGS` whereas `/` is
part of most filesystem paths. An alternative is to escape any usage of
`/` in the contents of `$DFLAGS`.

Fixes sociomantic-tsunami/makd#111.

### DIFF
--- a/mkversion.sh
+++ b/mkversion.sh
@@ -95,7 +95,7 @@ sed -i "$tmp" \
     -e "s/@DATE@/$date/" \
     -e "s/@AUTHOR@/$author/" \
     -e "s/@COMPILER@/$compiler/" \
-    -e "s/@DFLAGS@/$DFLAGS/" \
+    -e "s\#@DFLAGS@\#$DFLAGS\#" \
     -e "s/@FLAVOUR@/$F/"
 
 # Generate the libraries info


leandro-lucarella-sociomantic:
Mmm, the replacing is quite weak anyway, I guess it would be better to add a more robust replacing...

leandro-lucarella-sociomantic:
Thanks for the PR @jens-mueller-sociomantic, but I made #113 to replace it, as this is a bug fix, it should first go to the v1.11.x branch, which is the lowest maintained branch that had the issue. But actually #111 is showing a more general weakness in the current replacement scheme, so I prefer to fix the issue for all values, not just this one, so I will use `awk` to replace all values (making this a bug fix for v1.10.x, the lowest maintained branch).

If you agree with this, please mark it as "duplicate of #113" and close this :)

jens-mueller-sociomantic:
Duplicate of #113.

PR #115, opened 2017-12-08
Add coverage support

leandro-lucarella-sociomantic:
When using `make COV=1` the code will be compiled with `-cov`, and will
generate coverage reports when running.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -419,6 +419,9 @@ Variables you might want to override
   running the command to build D targets (when using the ``build_d`` function).
   By default they are used to generate the ``Version.d`` file, but users can
   override it not to generate the file or do something else on top of that.
+* ``COV`` will compile and run tests with coverage support if is set to ``1``. Please see Coverage_ for details.
+* ``COVDIR`` specifies the directory where to store coverage reports (by default ``$O/cov``. Please see Coverage_ for details.
+* ``COVMERGE`` indicates if coverage reports should be merged (``1`` will merge, ``0`` will not). Please see Coverage_ for details.
 
 Some of this variables are typically overridden in the Config.mak_ file, others
 in the Build.mak_ file, others in the Config.local.mak_ or directly in the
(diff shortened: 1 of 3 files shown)


leandro-lucarella-sociomantic:
Should be completed with the new runtime options to merge coverage reports.

leandro-lucarella-sociomantic:
Updated to match the current state of affairs in terms of runtime passing options.

leandro-lucarella-sociomantic:
Test PR: https://github.com/sociomantic-tsunami/ocean/pull/452

leandro-lucarella-sociomantic:
Updated to use `--DRT-covopt` for upstream DMD, as using env vars to configure the runtime is disabled by default (each program needs to enable it explicitly, in code).

leandro-lucarella-sociomantic:
This seems to work as expected judging from the ocean test PR. @mihails-strasuns-sociomantic do you want to try this out with some other project?

leandro-lucarella-sociomantic:
dmd-transitional support is due:
- [x] https://github.com/sociomantic-tsunami/dmd-transitional/pull/16
- [x] https://github.com/sociomantic-tsunami/dmd-transitional/pull/15

joseph-wakeling-sociomantic:
Is it intended that the generated files have leading `..` and `.` in their name?  This makes them hidden, which probably isn't what is wanted (I would guess it's a quirk of how the filenames are generated).

There's also a small discrepancy between what happens with stuff in the `src` and `submodules` directory (which wind up having filenames like `..src.whatever` and `..submodules.whatever`) and D source files that are auto-generated during the build process (which result in coverage files named `.home.username.path.to.project.build.whatever`.

FWIW other than that it seems to work OK.

mihails-strasuns-sociomantic:
@joseph-wakeling-sociomantic this is what DMD does, we don't have any control over that AFAIK

joseph-wakeling-sociomantic:
@mihails-strasuns-sociomantic no worries.  May be worth documenting in README and/or release notes, though.

leandro-lucarella-sociomantic:
Update compiler support.

README.rst
```diff
@@ -1170,3 +1173,31 @@ be re-compiled in that case!
...
+Coverage
+--------
+
+Compiling using code coverage can be done by passing ``COV=1`` to ``make``. If
+the D runtime supports the ``DRT_COV*`` environment variables (see list of
+version below), the coverage reports will be put in the ``$(COVDIR)`` directory
+(``$O/cov`` by default), otherwise they will be written in the top-level
+directory.
+
+Also by default the the coverage reports will be merged. To change this you can
```
joseph-wakeling-sociomantic, line 1187: What does it mean that the reports will be merged?  Might be worth putting in an explanation.

Going by what follows I assume that this refers to what happens if one builds multiple times in a row without cleaning?

joseph-wakeling-sociomantic (approved):
Still works fine for me, although I haven't tested anything related to the compiler support (I take it that was just an update to the README?).

leandro-lucarella-sociomantic:
> Still works fine for me, although I haven't tested anything related to the compiler support (I take it that was just an update to the README?).

Yes.

mihails-strasuns-sociomantic:
With this PR some of output still goes to root repo folder for both dmd1 and dmd-transitional. Same files are also present in cov tmp dir, so probably some binary target is simply missing required flags.

UPDATED: nevermind, it was simply a problem that `make clean` doesn't clean lst files not starting with `.` so those were leftover from old build.

leandro-lucarella-sociomantic:
OK, codecov is not working. I'm tired of waiting for it to confirm. Thanks to @joseph-wakeling-sociomantic I realized we might have an [issue](https://github.com/sociomantic-tsunami/makd/issues/121) though, but it would be a breaking change to fix it, so I'm delaying it for v3.

I guess as a first iteration of the new feature, we are fine.

PR #118, opened 2018-01-09
Use a new scheme for Debian versions

### LINKED ISSUE #117: Debian package versioning broken with pre-releases
# Part 1

Makd translate project's version to debian package compatible versions and as part of this process, a *debian_revision* is added with the distribution name, gor example:
`v1.2.1 -> 1.2.1-xenial`

The problem with this scheme is it was done before we actually started using pre-releases, which include a hyphen (`-`) too. So if we add a pre-release part to that example, we end up with:
`v1.2.1-rc.1 -> 1.2.1-rc.1-xenial`

Hyphens are special in [Debian version](https://www.debian.org/doc/debian-policy/#version). **Very** special. If only one is given, then everything after the hyphen is considered a *debian_revision*, but if more than one is given, then all of it is considered part of the *upstream_version*.

So the version string will be parsed differently by Debian for both examples:

1. `1.2.1-xenial`

   * *upstream_version*: `1.2.1`
   * *debian_revision*: `xenial`

2. `1.2.1-rc.1-xenial`

   * *upstream_version*: `1.2.1-rc.1-xenial`
   * *debian_revision*: (empty)

Because of this and how Debian parse versions (splitting in steps all non-numeric chars and comparing those lexicographically and comparing the numeric parts numerically), we end up with a *rc* version that is considered newer than a final version:

```console
$ if dpkg --compare-versions 1.2.1-xenial gt 1.2.1-rc.1-xenial; then echo YAY; else echo DAMN; fi
DAMN
```

Debian uses tilde (`~`) in a way that is similar to pre-releases or build metadata in semver. Basically `1.2.1~whatever` will be always older than `1.2.1`, and `1.2.1~~whatever` will be always even older than `1.2.1~~whatever`. According to Debian docs:

> The lexical comparison is a comparison of ASCII values modified so that all the letters sort earlier than all the non-letters and so that a tilde sorts before anything, even the end of a part. For example, the following parts are in sorted order from earliest to latest: `~~`, `~~a`, `~`, the empty part, `a`.

So the obvious fix is to just replace the hyphen with tilde (we are already using tilde widely):

```console
$ if dpkg --compare-versions 1.2.1-xenial gt 1.2.1~rc.1-xenial; then echo YAY; else echo DAMN; fi
YAY
```

This is easy except that there is also the versions produced from untagged versions.

# Part 2

Untagged versions (versions produced from a non-tagged commit or from a dirty repo) produce a version string that is a bit more complicated because it adds how many commits are we ahead from last tag. This is also already translated by Makd to a debian package compatible versions like this example:
`v1.2.1-rc.1-4-gea7105b-dirty -> 1.2.1~rc.1+4~dirty.20160202160552~ea7105b-xenial`

In principle this looks good, but in reality a dirty package should be newer than the same non-dirty version (as is the tagged version with some changes) so even when is in reality not related to this, we should fix that, just changing the tilde for a plus before the `dirty` should do it:
```console
$ if dpkg --compare-versions 1.2.1~rc.1+4~dirty.20160202160552~ea7105b-xenial gt 1.2.1~rc.1+4~ea7105b-xenial; then echo YAY; else echo DAMN; fi
DAMN
$ if dpkg --compare-versions 1.2.1~rc.1+4+dirty.20160202160552~ea7105b-xenial gt 1.2.1~rc.1+4~ea7105b-xenial; then echo YAY; else echo DAMN; fi
YAY
```

By adjusting just this bit, all seems to work as it should with Debian versions. Here is a simple test bed to verify this:

```sh
compare_all()
{
    while true
    do
        v=$1; shift
        test -n "$1" || 
            break
        for op in eq gt lt
        do
            dpkg --compare-versions "$v" $op "$1" && printf "%40s  %s  %s\n" "$v" $op "$1"
        done
    done
}
$ compare_all \
    1.8.3 1.8.3~rc.2+dirty.20180105195104 \
    1.8.3~rc.2 \
    1.8.3~rc.1+1+dirty.20180105195104~67cf \
    1.8.3~rc.1+1~67cf \
    1.8.3~rc.1 \
    1.8.2+4+dirty.20180105195323~abcf \
    1.8.2+4+dirty.20180105195104~abcf \
    1.8.2+4~abcf \
    1.8.2+2~123f \
    1.8.2 \
    1.8.2~rc.1+1~1234 \
    1.8.2~rc.1 \
    1.8.2~alpha.2+3~1234 \
    1.8.2~alpha.2 \
    1.8.2~alpha.1+1~1235 \
    1.8.2~alpha.1 \
    1.8.0+10~abfe \
    1.8.0
```

Results are:
```
                                   1.8.3  gt  1.8.3~rc.2+dirty.20180105195104
         1.8.3~rc.2+dirty.20180105195104  gt  1.8.3~rc.2
                              1.8.3~rc.2  gt  1.8.3~rc.1+1+dirty.20180105195104~67cf
  1.8.3~rc.1+1+dirty.20180105195104~67cf  gt  1.8.3~rc.1+1~67cf
                       1.8.3~rc.1+1~67cf  gt  1.8.3~rc.1
                              1.8.3~rc.1  gt  1.8.2+4+dirty.20180105195323~abcf
       1.8.2+4+dirty.20180105195323~abcf  gt  1.8.2+4+dirty.20180105195104~abcf
       1.8.2+4+dirty.20180105195104~abcf  gt  1.8.2+4~abcf
                            1.8.2+4~abcf  gt  1.8.2+2~123f
                            1.8.2+2~123f  gt  1.8.2
                                   1.8.2  gt  1.8.2~rc.1+1~1234
                       1.8.2~rc.1+1~1234  gt  1.8.2~rc.1
                              1.8.2~rc.1  gt  1.8.2~alpha.2+3~1234
                    1.8.2~alpha.2+3~1234  gt  1.8.2~alpha.2
                           1.8.2~alpha.2  gt  1.8.2~alpha.1+1~1235
                    1.8.2~alpha.1+1~1235  gt  1.8.2~alpha.1
                           1.8.2~alpha.1  gt  1.8.0+10~abfe
                           1.8.0+10~abfe  gt  1.8.0
```

Changing the scheme to this might prove tricky though, unless we keep assuming a lot about what an *upstream_version* could be (which we probably should anyway, otherwise versions won't work), but it seems to fix the issue.

# Pending Issues

When talking about non tagged versions, there is also the problem with semver. Unfortunately I don't think there is a way to map this behaviour to semver because I couldn't find any way to express v1.8.2+3 commits for example. Semver requires all X Y and Z in vX.Y.Z to be numbers. If we were to use `-` as a separator, then v1.8.3-3 would be older than v1.8.3 which is not what we want. If we use metadata separator, it doesn't have an effect on version ordering.

I guess this is not a big issue, as semver is not really for deployment but for dependency declaration and being so it should only refer to released tags. But is something to have in mind. Dirty information could be added as build metadata but it isn't exactly what we want either.

I guess we should just live with versions like `v1.2.1-rc.1-4-gea7105b-dirty` not being really semver compatible.

/cc @mihails-strasuns-sociomantic @nemanja-boric-sociomantic  @gavin-norman-sociomantic

leandro-lucarella-sociomantic:
Debian versions used by MakD are broken when the git tag is a pre-release as per SemVer. The problem is the `-` separator for SemVer pre-releases is not translated so, for example, version `1.2.3-xenial` is older than `1.2.3-alpha.1-xenial`, because by having 2 `-` in the version, then the whole string is interpreted as the upstream version, and `-` loses any special meaning.

To fix this the new scheme is used that should solve all the issues. Also `mkversion.sh` is adjusted to print untagged versions as `<branch>-0-g<H>` instead of the old `<branch>-g<H>`. This is to simplify parsing when converting to Debian versions.

This new scheme is as follows:

The input version is taken from `STDIN` and is expected to be in the form:
```
[v]<upstream_version>[-<C>-g<H>][-dirty] (as constructed by `mkversion.sh`).
```

Where:

- `<upstream_version>`
  Is the current version of the project, expected to be the last tag, which should be SemVer compatible, or a branch name.
  It will be parsed as a SemVer version:

  ```
  <X>.<Y>.<Z>[-<pre_release>][+<metadata>]
  ```
  If the parsing is not successful, `<X>`, `<Y>` and `<Z>` are interpreted as 0 and the <pre_release> is set to `git.<upstream_version>` (so is interpreted as a 0.0.0 pre-release). In this case any non-alphanumeric character that is not compatible with the SemVer character set allowed for pre-releases is replaced by `.`.

  In any case `-` chars in `<pre_release>` and `<metadata>` is translated to `.` (as `-` has a special meaning for Debian versions).

- `<C>`
  Is how many commits ahead of the last tag the project is.

- `<H>`
 Is the hash of the current version's commit.

Parts between `[]` are optional.

The translated version is then constructed as:
```
<X>.<Y>.<Z>[~<pre_release>][+<C>][+dirty.<date>][~<H>][~<metadata>]-<dist>
```

Where:

- `<date>` is the current date in `YYYYmmddHHMMSS` format.
- `<dist>` is the current Debian version codename (as returned by `lsb_release`).

Fixes #117.


leandro-lucarella-sociomantic:
I want to add some test for this (even when is only for this, it would be a start). But in the meantime I would appreciate testing with real projects.

leandro-lucarella-sociomantic:
Updated:

* Fixed an invalid char replacement (`-` was originally converted to `_` but that's not really a valid char for Debian packages either, so using `.` instead)
* Added tests. **YES! MAKD WILL AT LEAST HAVE SOME TESTS!!!** Maybe more will come in the future ( @mathias-lang-sociomantic you might like this one!)
* Added travis + beaver for automatic testing.

leandro-lucarella-sociomantic:
:green_apple: :green_book: :green_heart: :green_salad: :frog: :white_check_mark:

leandro-lucarella-sociomantic:
Test PR for tango: https://github.com/sociomantic-tsunami/tangort/pull/33

leandro-lucarella-sociomantic:
Mmm, there seems to be 2 `-xenial`... :unamused: 
https://travis-ci.org/leandro-lucarella-sociomantic/tangort/jobs/327299019#L510

nemanja-boric-sociomantic:
https://github.com/sociomantic-tsunami/makd/blob/31b514f23d1b8de17559ca668b8bce330c9faeca/Makd.mak#L152

leandro-lucarella-sociomantic:
Yeah, that's not the problem actually, but:

https://github.com/sociomantic-tsunami/makd/blob/31b514f23d1b8de17559ca668b8bce330c9faeca/Makd.mak#L151

leandro-lucarella-sociomantic:
I will just make `gitver2deb` not output the `-DIST` part.

leandro-lucarella-sociomantic:
Too bad that this problem wasn't actually detected by the tests... :man_facepalming: 

But sadly for that we need more comprehensive tests (one that produces packages and test the version of the resulting package).

leandro-lucarella-sociomantic:
Updated.

leandro-lucarella-sociomantic:
Now it looks good: https://travis-ci.org/leandro-lucarella-sociomantic/tangort/builds/327307066#L501

leandro-lucarella-sociomantic:
ping @mihails-strasuns-sociomantic @nemanja-boric-sociomantic 

Maybe @joseph-wakeling-sociomantic or @mathias-lang-sociomantic can help with some testing?

leandro-lucarella-sociomantic:
ping? :sob:

mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Seems to work as expected for me, haven't tried with any existing app though

joseph-wakeling-sociomantic:
Regarding the `dirty` component: will this affect D2 builds of packages, which rely on D2 conversion?  Or is the version number constructed before the conversion is done?

joseph-wakeling-sociomantic:
I've just tested, and indeed for D2 package builds we now get a horrible `dirty` in the version number.

This is going to be an annoyance for multiple reasons.

joseph-wakeling-sociomantic (changes requested):
I'd like to request that we block this PR until some more fine-grained concerns over this approach have been discussed.

.travis.yml
```diff
@@ -0,0 +1,31 @@
+# We will use docker to set up out environment, so don't use any particular
```
gautam-kotian-sociomantic: `our environment`?

leandro-lucarella-sociomantic: Woops! This typo is probably in all our D-related open source projects :man_facepalming:

leandro-lucarella-sociomantic:
> I'd like to request that we block this PR until some more fine-grained concerns over this approach have been discussed.

Hi Joe, this PR changes **nothing** about **when** the dirty information it is printed, it just changes the format. I see your point but that's a different issue, and one that already exist. This PR is fixing broken ordering for Debian package versions with pre-releases and we need to roll it out soon.

Can you test the changes in this PR? Can you verify the problem you mention exist in the current makd version and create a separate issue about that if it does?

In any case I think I will merge this today if I don't get any new feedback, as this issue is nasty, it makes `apt` consider pre-releases more recent than the final release.

.travis.yml
leandro-lucarella-sociomantic: Updated.

joseph-wakeling-sociomantic (approved):
OK, happy to remove objections in the circumstances.  I'll follow up and confirm if there is an outstanding issue w.r.t. `dirty` package builds.

joseph-wakeling-sociomantic:
In short:

* `dirty` does show up in local D2 builds for existing stable `makd` release (v2.1.2) but this does _not_ show up for packages built in our CI system

* this suggests there's something different about how our CI carries out the builds, compared to the separate `make d2conv` and `make pkg` calls one might typically make locally

* if one e.g. does `make DVER=2 PKG_SUFFIX=-d2 F=production d2conv pkg` then the "clean" version shows up, so I presume this is what CI is doing.

With this in mind, objections removed.  Looks like a user problem, not a design problem ;-)

mihails-strasuns-sociomantic:
> if one e.g. does make DVER=2 PKG_SUFFIX=-d2 F=production d2conv pkg then the "clean" version shows up, so I presume this is what CI is doing.

Repeating what I have said in Slack - this is actually a MakD bug (does not re-evaluating git based variables) as `make a; make` should always be equivalent to `make a b` with a well-formed dependencies. But looks like we actually need this bug very much right now so it should not be fixed :)

leandro-lucarella-sociomantic:
May the force be with us...

joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic just for the record, is it intended that in the case of a branch that's 2 commits after a tag like `v1.0.0+s1-alpha.1`, we should end up with a deb version along the lines of,

```
1.0.0+2~6548~s1.alpha.1-xenial
```

... ?  (i.e. 2 entries with a `~` separator)

See: https://travis-ci.org/sociomantic-tsunami/mxnet/builds/329501191 for the concrete example ;-)

mihails-strasuns-sociomantic:
I am not sure if `v1.0.0+s1-alpha.1` is well-formed by SemVer - https://semver.org doesn't explicitly mention if there is a specific expected order between pre-release version and metadata, but example only use `v1.0.0-alpha.1+s1` form and that makes more sense in general considering metadata is ignored when ordering and pre-release version is not.

joseph-wakeling-sociomantic:
@mihails-strasuns-sociomantic gotcha.  In this case I could probably delete the release tag since there is nothing official yet ;-)

joseph-wakeling-sociomantic:
However, I would note that the usage is intended.  It's the alpha version of _our_ release; the underlying `v1.0.0` of the upstream project is stable.

mihails-strasuns-sociomantic:
It can be part of metadata too, like `v1.0.0+s1.alpha.1`. Note that from the SemVer PoV it will be completely ignored as anything that is stored in metadata (after `+`) is treated as raw string. And that makes sense because you can't keep SemVer promises if you have two different versions merged into one.

joseph-wakeling-sociomantic:
@mihails-strasuns-sociomantic sure, although I'm not sure it makes much practical difference in the current schema whether I tag `v1.0.0+s1-alpha1` or `v1.0.0+s1.alpha.1` (inasmuch as, the deb version comes out the same either way).

The main concerns from my side are:

* are the double `~` entries problematic for the deb package version?

* will our package upload mechanisms understand that this is a prerelease?

leandro-lucarella-sociomantic:
> are the double ~ entries problematic for the deb package version?

No, for Debian checks the version incrementally always using the same algorithm, and `~` only mean it would be considered older and anything else. It's better to RTFM as explaining it is not that easy.

> will our package upload mechanisms understand that this is a prerelease?

Yes, where to upload packages is determined by the Git **tag**, not the debian version.

joseph-wakeling-sociomantic:
So in the case where the tag is `v1.0.0+s1-alpha.1`, how will this be treated w.r.t. package upload?

leandro-lucarella-sociomantic:
I think it's considered a prerelease if `-` is present, but I agree with @mihails-strasuns-sociomantic that probably `v1.0.0+s1-alpha.1` is not a valid SemVer version, so I would stick to SemVer valid versions:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers **immediately following the patch version**.

https://semver.org/#spec-item-9

> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately **following the patch or pre-release version**.

https://semver.org/#spec-item-10

(emphasis by me)

joseph-wakeling-sociomantic:
I see.  So we're currently tied to needing to use something like `v1.0.0-alpha.1+s1` even though the alpha is on our side, rather than upstream?

Is there any chance we could address that?  It seems important/relevant to be able to distinguish between "alpha.1 release of sociomantic release 1 of upstream release v1.0.0" compared to "sociomantic release 1 of upstream release v1.0.0-alpha.1".

I'll note that when originally discussing the whole `.sN` or `+sN` idea originally, the `+sN-rc.1` etc. style of versioning seemed to be encouraged/approved -- see all our existing releases:
https://bintray.com/sociomantic-tsunami/mxnet/libmxnet

joseph-wakeling-sociomantic:
As a compromise I suppose we could use something like `v1.0.0-s1.alpha.1` for pre-releases and `v1.0.0+s1` for final releases ... ?  I'll try that out and see if it works.

leandro-lucarella-sociomantic:
> I see. So we're currently tied to needing to use something like v1.0.0-alpha.1+s1 even though the alpha is on our side, rather than upstream?

Yes, sadly SemVer doesn't have the concept of an "upstream version" as Debian does

> .Is there any chance we could address that?

I don't see any way to do that with SemVer. Not sure if it's worth taking to Neptune (which would break compatibility with SemVer, unless all we do is convey more meaning to some already existing SemVer component).

> It seems important/relevant to be able to distinguish between "alpha.1 release of sociomantic release 1 of upstream release v1.0.0" compared to "sociomantic release 1 of upstream release v1.0.0-alpha.1".

Maybe just prepend `s` to the pre-release for sociomantic pre-releases? `v1.0.0-s.alpha.1+s1` is Sociomantic's alpha 1 for upstream v1.0.0 and `v1.0.0-alpha.1+s1` is Sociomantic's release for upstream alpha 1? This will have the effect to make sociomantic pre-releases always be considered more recent than any upstream release, which might be a good thing (?)

> I'll note that when originally discussing the whole .sN or +sN idea originally, the +sN-rc.1 etc. style of versioning seemed to be encouraged/approved -- see all our existing releases:

I don't remember discussing `+sN` (but I'm not saying it didn't happen :). If it was discussed and agreed on, I guess we missed this detail. I remember saying that there was no SemVer way to express our releases based on upstream releases.

> As a compromise I suppose we could use something like v1.0.0-s1.alpha.1 for pre-releases and v1.0.0+s1 for final releases ... ? I'll try that out and see if it works.

I swear I wrote my proposal before reading this. If we independently reached the same conclusion, I guess it's a good sign :)

Please always separate letters and numbers with `.` in semver so you don't screw the order if the number goes beyond one digit.

joseph-wakeling-sociomantic:
> Maybe just prepend `s` to the pre-release for sociomantic pre-releases?

I think it'll need to be something like `-s1.alpha.1` and `-s2.alpha.1` etc. because otherwise, how does one distinguish between alpha 1 of sociomantic release 1, and alpha 1 of sociomantic release 2?

(This is a real use-case, e.g. if we don't update the underlying project release, but we do update something about our packaging setup.)

joseph-wakeling-sociomantic:
> I swear I wrote my proposal before reading this. If we independently reached the same conclusion, I guess it's a good sign :)

... and I wrote my response before reading this far ... :-P

> Please always separate letters and numbers with . in semver so you don't screw the order if the number goes beyond one digit.

So you mean `-s.1.alpha.1` instead of `-s1.alpha.1` ... ?

joseph-wakeling-sociomantic:
> I don't remember discussing `+sN` (but I'm not saying it didn't happen :). If it was discussed and agreed on, I guess we missed this detail. I remember saying that there was no SemVer way to express our releases based on upstream releases.

Yea, it was a while back.  I had to read back into our discussions some way just to get some idea of what we agreed.

How about I write up a little piece of documentation on this use case?  Then we can finalize the discussion and the proposal.  Would it be best submitted to the neptune docs, or somewhere else?

joseph-wakeling-sociomantic:
For reference, a version tag like `v1.0.0-s.1.alpha.1` produces a deb version of:

```
1.0.0~s.1.alpha.1-xenial
```

leandro-lucarella-sociomantic:
> I think it'll need to be something like -s1.alpha.1 and -s2.alpha.1 etc. because otherwise, how does one distinguish between alpha 1 of sociomantic release 1, and alpha 1 of sociomantic release 2?

The problem with this is ordering, `-s2.alpha.1` will be newer than `-s1.rc.23`, which makes no sense.

leandro-lucarella-sociomantic:
> How about I write up a little piece of documentation on this use case? Then we can finalize the discussion and the proposal. Would it be best submitted to the neptune docs, or somewhere else?

Lets first get something written down and after seeing how it looks we see where to put it.

joseph-wakeling-sociomantic:
> The problem with this is ordering, `-s2.alpha.1` will be newer than `-s1.rc.23`, which makes no sense.

But it does make sense -- alpha 1 of sociomantic release 2 _should_ be considered newer than release candidate 23 of sociomantic release 1.  No ... ?

leandro-lucarella-sociomantic:
Oh, OK, I see, I guess I got a bit lost in the way. So OK for that case, you are right. But I'm not sure if that's a valid use case. You are basically just packaging upstream, right, or are you making changes too? OK, in any case maybe it's better to consider the more general case, as we have some (like ebtree).

So now upstram released `v1.2.3-alpha.1`, you make some changes to it and want to do a release for that. Let's say you are confident and want to make a release candidate for the first internal release. So that would be... `v1.2.3-alpha.1.s.1.rc.1`? I feel like it is getting too meta. The version is already an alpha version in any case, but I guess you could do that.

I also wonder if the `s` will be hidden in the middle of other info if we should go back to `socio` instead, as it's more clear.

In any case, I feel this is getting too off-topic here. I guess it might be better to just create an issue in neptune. Ideally with a proposal that considers all cases and make sure the ordering is right.

joseph-wakeling-sociomantic:
OK, I'll raise a neptune issue and we can continue the discussion there.  FWIW I agree about the idea of `socio` instead of just `s` (I would even go so far as to say `sociomantic`).

I'd like to get this tied down soon if possible because we'd like to roll out a new `libmxnet` package release in the near future.

joseph-wakeling-sociomantic:
Neptune issue: https://github.com/sociomantic-tsunami/neptune/issues/145

joseph-wakeling-sociomantic:
Just to note, since I've been reading through the upstream documentation:

> The problem is the - separator for SemVer pre-releases is not translated so, for example, version 1.2.3-xenial is older than 1.2.3-alpha.1-xenial, because by having 2 - in the version, then the whole string is interpreted as the upstream version, and - loses any special meaning.

This seems at odd with what's stated in the deb versioning guide when discussing the `debian_revision` (https://www.debian.org/doc/debian-policy/#version):

> The package management system will break the version number apart at the last hyphen in the string (if there is one) to determine the `upstream_version` and `debian_revision`.

In other words, a package whose version is:

```
1.0.0-alpha.1-xenial
```

... ought to have upstream version `1.0.0-alpha.1` and debian revision `xenial`.  So if apt or dpkg are interpreting `1.0.0-alpha.1-xenial` as more recent than `1.0.0-xenial`, that suggests something different might have been in play.

leandro-lucarella-sociomantic:
mmm, I guess I read this as "if there is ...":

> If there is no debian_revision then hyphens are not allowed.

In any case, `1.0.0-alpha.1` is considered newer than `1.0.0` as an upstream version:
```console
$ dpkg --compare-versions 1.0.0-alpha.1 gt 1.0.0 && echo YES
YES
```

It should be `1.0.0~alpha.1` to make it work as expected.

PR #120, opened 2018-02-01
Add support for specifying a custom package iteration

joseph-wakeling-sociomantic:
Package versioning includes at least two components:

  * the upstream package version (e.g. `1.0.3`)

  * the package iteration (e.g. `0ubuntu0`)

which for deb packages result in a package version like:

    1.0.3-0ubuntu0

Makd has until now hardcoded the assumption that the package iteration should be the same as the distro codename (`trusty`, `xenial`, etc.), but this is not typical practice in Debian packaging, where it is used instead to indicate detailed information on the version of the package (rather than the project being packaged); `0ubuntu0`, `1ubuntu3`, etc.

This patch adds a new, customizable `PKGITERATION` variable which can be set in build scripts or via the command line.  It defaults to the distro codename (meaning nothing will change if it is not explicitly set), but allows the user to override this default.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -415,6 +415,9 @@ Variables you might want to override
   obtained via the ``mkversion.sh`` by default.
 * ``PKGVERSION`` is the version to be used when creating packages. It's
   obtained via the ``VERSION`` variable by default.
+* ``PKGITERATION`` optionally allows setting the package *iteration* (forwarded
+  to `fpm` as `--iteration`, known as the `debian_revision`
+  [in Debian](https://www.debian.org/doc/debian-policy/#version)).
 * ``PRE_BUILD_D`` and ``POST_BUILD_D`` hold scripts executed before and after
   running the command to build D targets (when using the ``build_d`` function).
   By default they are used to generate the ``Version.d`` file, but users can
@@ -764,6 +767,9 @@ passed to the ``mkpkg`` util. By default Makd passes the following variables:
 
 ``version``
         package version number as defined by ``PKGVERSION``.
+
+``iteration``
+        package iteration as defined by ``PKGITERATION``.
 
 ``builddir``
         base build directory (``$G``).
--- a/relnotes/pkg-iteration.feature.md
+++ b/relnotes/pkg-iteration.feature.md
@@ -1,0 +1,25 @@
+## The package iteration can now be specified via `PKGITERATION`
+
+Package versions consist of two components: the _upstream version_, and
+the _package iteration_ (corresponding to the `debian_revision` where
+deb packages are concerned).
+
+Previously the package iteration was hardcoded to match the distro code
+(e.g. `xenial`), disallowing typical patterns referencing the packager
+and changes to the packaging rather than the upstream project.
+
+The new `PKGITERATION` variable allows the package iteration to be set
+to whatever is wanted, whether via `Config.mak` or the command line.
+It still defaults to the distro code, meaning nothing will change for
+existing package definitions, but can now be customized where this is
+appropriate (e.g. for packages of 3rd-party projects).
+
+Note that if you want to include the distro code in a customized package
+iteration, this will have to be included manually.  For example,
+`PKGITERATION=3~"$(lsb_release -cs)"` will produce iterations such as
+`3~trusty`, `3~xenial`, etc. depending on the distro used to build the
+package.
+
+For some examples of how to set the package iteration, see the
+ [Debian versioning policy](https://www.debian.org/doc/debian-policy/#version)
+and [the Ubuntu `debian_revision` schema](https://askubuntu.com/questions/620533/how-does-ubuntu-name-packages).
--- a/Makd.mak
+++ b/Makd.mak
@@ -146,9 +146,13 @@ PKG_FILES ?= $(wildcard $(PKG)/*.pkg)
 # (defined lazily so we can use target variables)
 ifndef PKG_DEFAULTS
 DISTRO_CODENAME="$(shell lsb_release -rs | grep rolling || lsb_release -cs)"
+
+# Default package iteration
+PKGITERATION ?= "$(DISTRO_CODENAME)"
+
 PKG_DEFAULTS = -D-f -D-sdir -D-t$(PKGTYPE) -D--deb-changelog="$O/changelog.Debian" \
 		-D--version="$(PKGVERSION)" \
-		-D--iteration="$(DISTRO_CODENAME)" \
+		-D--iteration="$(PKGITERATION)" \
 		-d lsb_release="$(DISTRO_CODENAME)"
 endif
 
@@ -486,6 +490,7 @@ $O/pkg-%.stamp: $(PKG)/%.pkg $G/pkg-flags
 		$(if $V,,-vv) -D-p"$P" -F "$(FPM)" \
 		$(PKG_DEFAULTS) \
 		-d suffix="$(PKG_SUFFIX)" -d version="$(PKGVERSION)" \
+		-d iteration="$(PKGITERATION)" \
 		-d builddir="$G" -d bindir="$B" \
 		-d fullname="$*$(PKG_SUFFIX)" -d shortname="$*" \
 		$<,$<,mkpkg)


joseph-wakeling-sociomantic:
@leandro-lucarella-sociomantic I think this should allow us to better handle custom deb versions.  Just as a note: I think it's worth abandoning the notion of "alpha/beta/rc package versions" discussed previously.  Just bumping the package iteration should suffice.

Whether the package ends up in a prerelease repo or a release repo, of course, is a separate concern and should be handled separately.

The current way of doing things can be handled by e.g.

```
make PKGVERSION=1.0.0 PKGITERATION=0tsunami0~"$(lsb_release -cs)"
```

to get a final version of `1.0.0-0tsunami0~xenial` (for example).

joseph-wakeling-sociomantic:
Note, it's deliberate to not _force_ the iteration to contain the distro codename.  While it's common practice for 3rd-party packagers (as opposed to in-distro packages), it's not universal.

leandro-lucarella-sociomantic (changes requested):
Can you also write a release note and edit the README to mention the new variable (both in the variables section and in the packaging section)?

joseph-wakeling-sociomantic:
Added release notes and edited the README:

```diff
diff --git a/README.rst b/README.rst
index db5bf3f..eeae579 100644
--- a/README.rst
+++ b/README.rst
@@ -415,6 +415,9 @@ Variables you might want to override
   obtained via the ``mkversion.sh`` by default.
 * ``PKGVERSION`` is the version to be used when creating packages. It's
   obtained via the ``VERSION`` variable by default.
+* ``PKGITERATION`` is the iteration of the package to be created.  By default
+  it is set to the distro code (e.g. ``trusty``, ``xenial``).  It will be
+  ignored if ``PKG_DEFAULTS`` is set manually.
 * ``PRE_BUILD_D`` and ``POST_BUILD_D`` hold scripts executed before and after
   running the command to build D targets (when using the ``build_d`` function).
   By default they are used to generate the ``Version.d`` file, but users can
@@ -765,6 +768,9 @@ passed to the ``mkpkg`` util. By default Makd passes the following variables:
 ``version``
         package version number as defined by ``PKGVERSION``.
 
+``iteration``
+        package iteration as defined by ``PKGITERATION``.
+
 ``builddir``
         base build directory (``$G``).
 
diff --git a/relnotes/pkg-iteration.feature.md b/relnotes/pkg-iteration.feature.md
new file mode 100644
index 0000000..147c4b2
--- /dev/null
+++ b/relnotes/pkg-iteration.feature.md
@@ -0,0 +1,15 @@
+## The package iteration can now be specified via `PKGITERATION`
+
+* `Makd.mak`
+
+  Package versions consist of two components: the upstream version, and
+  the package iteration.  Previously this was hardcoded to match the
+  distro code (e.g. `xenial`), disallowing typical patterns referencing
+  the packager and changes to the packaging rather than the upstream
+  project (e.g. `1ubuntu0`, `3debian1.2`).
+
+  The new `PKGVERSION` variable allows this to be set to whatever is
+  wanted for the package, whether via `Config.mak` or the command line.
+  Note that if you want to still include the distro code, this will have
+  to be included manually as part of the custom iteration choice, for
+  example: `PKGITERATION=0tsunami0~"$(lsb_release -cs)"`
```

joseph-wakeling-sociomantic:
About the milestone: no chance of getting this into the v2.2.0 release ... ?

joseph-wakeling-sociomantic:
Example use case in the mxnet deb-package makefile:
https://github.com/joseph-wakeling-sociomantic/mxnet/commit/5ab1b0acbd050e5c72021aacc5c13213964da1dd

No doubt this will look horrific from your point of view, but it works for the use case we have.  (Always happy to hear more robust suggestions of course!)

joseph-wakeling-sociomantic:
>  but it works for the use case we have

... or not.  I'll get back to this tomorrow ;-)

leandro-lucarella-sociomantic:
> About the milestone: no chance of getting this into the v2.2.0 release ... ?

Yes, I thought v2.3.0 was next release because there was no v2.2.0 milestone. I will fix the milestone.

README.rst
```diff
@@ -415,6 +415,9 @@ Variables you might want to override
   obtained via the ``mkversion.sh`` by default.
 * ``PKGVERSION`` is the version to be used when creating packages. It's
   obtained via the ``VERSION`` variable by default.
+* ``PKGITERATION`` is the iteration of the package to be created.  By default
+  it is set to the distro code (e.g. ``trusty``, ``xenial``).  It will be
+  ignored if ``PKG_DEFAULTS`` is set manually.
```
leandro-lucarella-sociomantic: This doesn't sound right, after all is `PKG_DEFAULTS`, not `PKG_OVERRIDES` or something like that. Would it be too hard to change this?

README.rst
```diff
@@ -765,6 +768,9 @@ passed to the ``mkpkg`` util. By default Makd passes the following variables:
 ``version``
         package version number as defined by ``PKGVERSION``.
 
+``iteration``
+        package iteration as defined by ``PKGITERATION``.
+
```
leandro-lucarella-sociomantic, line 773: You need to change `mkpkg` to expose `PKGITERATION`, is not automagically ✨

relnotes/pkg-iteration.feature.md
```diff
@@ -0,0 +1,15 @@
+## The package iteration can now be specified via `PKGITERATION`
+
+* `Makd.mak`
```
leandro-lucarella-sociomantic: Don't make this a list, the new format suggest just a h2 title and then regular paragraphs.

relnotes/pkg-iteration.feature.md
```diff
@@ -0,0 +1,15 @@
...
+
+  Package versions consist of two components: the upstream version, and
+  the package iteration.  Previously this was hardcoded to match the
```
leandro-lucarella-sociomantic: Since the main supported packaging is Debian, I think it might be worth to clarify that *package iteration* in Debian lingo is the `debian_version`.

leandro-lucarella-sociomantic (changes requested):

README.rst
joseph-wakeling-sociomantic: Hmmm.  Let's put it this way: I think I'd rather implement this first in a way that's in line with how Makd currently works, and then look separately at the question of whether Makd _should_ work this way.

For what it's worth, I think the existing behaviour is not unreasonable.  `PKG_DEFAULTS` is perhaps misnamed, but it's a catch-all collection of all the flags to be passed to `fpm`.  The only reason to override this is to make a completely custom set of `fpm` flags.

From this point of view, things like `PKGVERSION` AND `PKGITERATION` only make sense in the context of a predictable approach to how the `fpm` flags are generated.

On the other hand, an alternative approach could be to redefine things so that:

* some flags are always expected to be set (version and iteration being two obvious ones, but things like package type being another), and are overridable by individual variables like `PKGVERSION`, `PKGITERATION`, and `PKGTYPE`

  * the `fpm` flags could be omitted if the variable is unset or empty

* define `PKG_DEFAULTS` in terms of a combination of these obligatory flags, and a new `PKG_DEFAULT_FLAGS` variable to contain the fixed flags currently hardcoded into `PKG_DEFAULTS`.  There are actually only two that meet this definition: `-D-f -D-sdir`.

  * the `-D--deb-changelog=` flag is currently hardcoded, but it would be easy to add a variable to set it (note of course that this is a `deb`-specific setting).

In other words, instead of the current:

```mak
# Default fpm flags. By default it builds Debian packages from files, a Debian
# changelog is provided, and a version and iteration (using the Debian version)
# (defined lazily so we can use target variables)
ifndef PKG_DEFAULTS
DISTRO_CODENAME="$(shell lsb_release -rs | grep rolling || lsb_release -cs)"

# Default package iteration
PKGITERATION ?= "$(DISTRO_CODENAME)"

PKG_DEFAULTS = -D-f -D-sdir -D-t$(PKGTYPE) -D--deb-changelog="$O/changelog.Debian" \
        -D--version="$(PKGVERSION)" \
        -D--iteration="$(PKGITERATION)" \
        -d lsb_release="$(DISTRO_CODENAME)"
endif
```

we could have:

```mak
# Default fpm flags. By default it builds Debian packages from files, a Debian
# changelog is provided, and a version and iteration (using the Debian version)
# (defined lazily so we can use target variables)
ifndef PKG_DEFAULTS
DISTRO_CODENAME="$(shell lsb_release -rs | grep rolling || lsb_release -cs)"

# Default package iteration
PKGITERATION ?= "$(DISTRO_CODENAME)"

PKG_DEFAULT_FLAGS ?= -D-f -D-sdir -D--deb-changelog="$O/changelog.Debian"

PKG_DEFAULTS = -D-t$(PKGTYPE) \
        -D--version="$(PKGVERSION)" \
        -D--iteration="$(PKGITERATION)" \
        -d lsb_release="$(DISTRO_CODENAME)" \
        $(PKG_DEFAULT_FLAGS)
endif
```

This would not avoid the basic fact that overriding `PKG_DEFAULTS` would leave all other settings irrelevant, but it would make a better way possible without breaking change.  We could then strip out the ability to override `PKG_DEFAULTS` in the next major release.

joseph-wakeling-sociomantic: The TL;DR answer is: we can and should improve how makd works w.r.t. `PKG_DEFAULTS` but I don't think we should tie it to this PR.  Because any real fix involves a breaking change, and this PR's new feature doesn't need to.

README.rst
joseph-wakeling-sociomantic: Ah.  I assumed it would be handled automatically by this call:

```mak
    $(call exec,PYTHONDONTWRITEBYTECODE=1 $(MAKD_PATH)/mkpkg \
        $(if $V,,-vv) -D-p"$P" -F "$(FPM)" \
        $(PKG_DEFAULTS) \
        -d suffix="$(PKG_SUFFIX)" -d version="$(PKGVERSION)" \
        -d builddir="$G" -d bindir="$B" \
        -d fullname="$*$(PKG_SUFFIX)" -d shortname="$*" \
        $<,$<,mkpkg)
```

... where the `$(PKG_DEFAULTS)` are included, and therefore I assumed `iteration` would be included.  What do I need to change in `mkpkg` ... ?  (I'm afraid I'm having some difficulty following the python.)

relnotes/pkg-iteration.feature.md
joseph-wakeling-sociomantic: `debian_revision`, not debian version (see: https://www.debian.org/doc/debian-policy/#version).  But yup, good call.

joseph-wakeling-sociomantic:
I've added a fixup patch that should cover the requested tweaks to README and release notes (in particular, stressing the `debian_revision` connection), as well as correcting some errors.

I haven't addressed the issue that custom-setting `PKG_DEFAULTS` overrides settings like `PKGVERSION` and `PKGITERATION`.  As described above, this would be a breaking change so is probably best handled separately (although I can obviously change things if this is wanted).

I also haven't addressed the issue of `mkpkg` exposing `PKGITERATION` as `iteration`, just because I haven't followed what is wanted/needed.  See also my remarks above: https://github.com/sociomantic-tsunami/makd/pull/120#discussion_r165706109 (unless I need to simply add another `-d` flag to the statement referenced there?).

joseph-wakeling-sociomantic:
> I also haven't addressed the issue of mkpkg exposing `PKGITERATION` as `iteration`

@leandro-lucarella-sociomantic can I ping on the above please?  Would quite like to get this done, as other issues are blocked on it :-\

README.rst
leandro-lucarella-sociomantic: That would pass `-i` to `fpm`, but the `.pkg` script will know nothing about it. To *define* variables for `.pkg` scripts, you need to pass: `-d iteration="$(PKGITERATION)"` or something.

leandro-lucarella-sociomantic: Arguments to `fpm` != arguments to `mkpkg`.

relnotes/pkg-iteration.feature.md
```diff
@@ -0,0 +1,21 @@
...
+
+Previously the package iteration was hardcoded to match the distro code
+(e.g. `xenial`), disallowing typical patterns referencing the packager
+and changes to the packaging rather than the upstream project.  For
+example, Ubuntu packages typically use an `XubuntuY` format where `X`
+is the numerical revision of the Debian package that has been forked
+(or `0` if there is no such package), and `Y` is the Ubuntu package
+revision (hence Ubuntu package versions like `2.7.0-2ubuntu3.1`, where
+`2.7.0` is the upstream version and `2ubuntu3.1` means Ubuntu revision
+3.1 of Debian package revision 2).
```
leandro-lucarella-sociomantic: Not a huge deal, and I'm fine if you want to leave this, but I think the whole thing about how ubuntu is using encoding the `debian_revision` is irrelevant and distracting.

relnotes/pkg-iteration.feature.md
```diff
@@ -0,0 +1,21 @@
...
+
+The new `PKGITERATION` variable allows the package iteration to be set
+to whatever is wanted for the package, whether via `Config.mak` or the
+command line.  Note that if you want to still include the distro code,
+this will have to be included manually as part of the custom iteration
+choice, for example: `PKGITERATION=0tsunami0~"$(lsb_release -cs)"`.
```
leandro-lucarella-sociomantic: I would also use a more simple example, like `PKGITERATION=1` because unless you have the whole explanation about what each number represents there, it's just cryptic. And if you follow debian packaging guidelines, the examples will just have `1` as the debian revision. I have the feeling the whole thing about Ubuntu is just guiding users to an unnecessarily complex scheme.

joseph-wakeling-sociomantic:
Squashed, and updated to pass `iteration` to `mkpkg`:

```diff
diff --git a/Makd.mak b/Makd.mak
index be13f04..bf03ca3 100644
--- a/Makd.mak
+++ b/Makd.mak
@@ -490,6 +490,7 @@ $O/pkg-%.stamp: $(PKG)/%.pkg $G/pkg-flags
                $(if $V,,-vv) -D-p"$P" -F "$(FPM)" \
                $(PKG_DEFAULTS) \
                -d suffix="$(PKG_SUFFIX)" -d version="$(PKGVERSION)" \
+               -d iteration="$(PKGITERATION)" \
                -d builddir="$G" -d bindir="$B" \
                -d fullname="$*$(PKG_SUFFIX)" -d shortname="$*" \
                $<,$<,mkpkg)
```

joseph-wakeling-sociomantic:
About the `debian_revision` thing: it seems like a reasonably widespread standard (I found multiple sources using it, not just Ubuntu), so I didn't want to leave that use-case undocumented.  I'll tweak the last example, though.

joseph-wakeling-sociomantic:
I've tweaked the second example in the release notes:

```diff
diff --git a/relnotes/pkg-iteration.feature.md b/relnotes/pkg-iteration.feature.md
index 271ae15..b338a5b 100644
--- a/relnotes/pkg-iteration.feature.md
+++ b/relnotes/pkg-iteration.feature.md
@@ -18,4 +18,4 @@ The new `PKGITERATION` variable allows the package iteration to be set
 to whatever is wanted for the package, whether via `Config.mak` or the
 command line.  Note that if you want to still include the distro code,
 this will have to be included manually as part of the custom iteration
-choice, for example: `PKGITERATION=0tsunami0~"$(lsb_release -cs)"`.
+choice, for example: `PKGITERATION=3~"$(lsb_release -cs)"`.
```

leandro-lucarella-sociomantic:
> About the debian_revision thing: it seems like a reasonably widespread standard (I found multiple sources using it, not just Ubuntu), so I didn't want to leave that use-case undocumented. I'll tweak the last example, though.

Again, that only makes sense if you are forking a Debian package (Debian as in maintained by the Debian project), **not** when you are the original author of a package. See the [Debian New Maintainer's Guide](https://www.debian.org/doc/manuals/maint-guide/index.en.html). They recommend using `dch -i` to create a new version of the package you maintain, which basically will just increment the *debian_revision*. In all the examples there it uses just a number (1) for the `debian_revision`. I really don't see any reason to complicate things and keep wasting time in looking for people defining more complicated schemes when we don't need them. I don't think we should recommend that.

leandro-lucarella-sociomantic (commented):
OK, the more we discuss this *debian_revision* based on a package that doesn't exist the more I think we should remove all that explanation about the complex revision scheme used to fork packages from another author. Besides that, LGTM, but before merging I would like to know why it is justified to complicate things and distract users reading the release notes with this :)

joseph-wakeling-sociomantic:
> Again, that only makes sense if you are forking a Debian package (Debian as in maintained by the Debian project), not when you are the original author of a package.

Yes, but that guide is written for _people working internally within Debian_.

The `XwhateverY` format for `debian_revision` is designed explicitly for people providing `deb` packages _outside_ of the Debian project.  It explicitly covers the case of where one is providing an original package of a 3rd-party project (`X == 0`), rather than forking a Debian package.  And there are multiple examples of this in action: Ubuntu uses that scheme, but so does e.g. the git PPA (the current git version distributed there is `1:2.16.1-0ppa1~ubuntu16.04.1`), and so do others again.

What I _will_ do is add some more info to the release notes indicating that a simple revision number is a perfectly acceptable choice for the `debian_revision`.

joseph-wakeling-sociomantic:
I've updated the release notes to give a more general flavour of the options and use-cases, putting the simple numerical revision first:

```diff
diff --git a/relnotes/pkg-iteration.feature.md b/relnotes/pkg-iteration.feature.md
index b338a5b..c215952 100644
--- a/relnotes/pkg-iteration.feature.md
+++ b/relnotes/pkg-iteration.feature.md
@@ -7,15 +7,32 @@ deb packages are concerned).
 Previously the package iteration was hardcoded to match the distro code
 (e.g. `xenial`), disallowing typical patterns referencing the packager
 and changes to the packaging rather than the upstream project.  For
-example, Ubuntu packages typically use an `XubuntuY` format where `X`
-is the numerical revision of the Debian package that has been forked
-(or `0` if there is no such package), and `Y` is the Ubuntu package
-revision (hence Ubuntu package versions like `2.7.0-2ubuntu3.1`, where
-`2.7.0` is the upstream version and `2ubuntu3.1` means Ubuntu revision
-3.1 of Debian package revision 2).
+example:
+
+* Debian packages use a simple number for the `debian_revision` (1, 2,
+  3, ...), resetting to 1 when the upstream version changes
+
+* Ubuntu packages typically use an `XubuntuY` format where `X` is the
+  numerical revision of the Debian package that has been forked (or `0`
+  if there is no such package), and `Y` is the Ubuntu package revision
+  (hence Ubuntu package versions like `2.7.0-2ubuntu3.1`, where `2.7.0`
+  is the upstream version and `2ubuntu3.1` means Ubuntu revision 3.1 of
+  Debian package revision 2)
+
+* some packagers like to include both revision information and the build
+  or target distro: e.g. packages from the git PPA have revisions such
+  as `0ppa1~ubuntu16.04.1` (note also the leading 0 indicating that this
+  is a new package rather than a fork of a Debian package).
 
 The new `PKGITERATION` variable allows the package iteration to be set
 to whatever is wanted for the package, whether via `Config.mak` or the
 command line.  Note that if you want to still include the distro code,
 this will have to be included manually as part of the custom iteration
-choice, for example: `PKGITERATION=3~"$(lsb_release -cs)"`.
+choice, for example: `PKGITERATION=3~"$(lsb_release -cs)"` to produce
+iterations of `3~trusty`, `3~xenial`, etc. depending on the distro used
+to build the package.
+
+See https://www.debian.org/doc/debian-policy/#version for the official
+Debian package-versioning spec, and
+https://askubuntu.com/questions/620533/how-does-ubuntu-name-packages
+for an explanation of how Ubuntu specifies the `debian_revision`.
```

joseph-wakeling-sociomantic:
Whoops, forgot to push some of the aforementioned changes.  Fixed now.

joseph-wakeling-sociomantic:
One more release-notes tweak to mention the default behaviour of the new variable:

```diff
diff --git a/relnotes/pkg-iteration.feature.md b/relnotes/pkg-iteration.feature.md
index c215952..beda1f3 100644
--- a/relnotes/pkg-iteration.feature.md
+++ b/relnotes/pkg-iteration.feature.md
@@ -25,12 +25,16 @@ example:
   is a new package rather than a fork of a Debian package).
 
 The new `PKGITERATION` variable allows the package iteration to be set
-to whatever is wanted for the package, whether via `Config.mak` or the
-command line.  Note that if you want to still include the distro code,
-this will have to be included manually as part of the custom iteration
-choice, for example: `PKGITERATION=3~"$(lsb_release -cs)"` to produce
-iterations of `3~trusty`, `3~xenial`, etc. depending on the distro used
-to build the package.
+to whatever is wanted, whether via `Config.mak` or the command line.
+It defaults to the distro code, meaning nothing will change for existing
+package definitions, but can now be customized where this is appropriate
+(e.g. for packages of 3rd-party projects).
+
+Note that if you want to include the distro code in a customized package
+iteration, this will have to be included manually.  For example,
+`PKGITERATION=3~"$(lsb_release -cs)"` will produce iterations such as
+`3~trusty`, `3~xenial`, etc. depending on the distro used to build the
+package.
 
 See https://www.debian.org/doc/debian-policy/#version for the official
 Debian package-versioning spec, and
```

relnotes/pkg-iteration.feature.md
```diff
@@ -0,0 +1,42 @@
...
+Note that if you want to include the distro code in a customized package
+iteration, this will have to be included manually.  For example,
+`PKGITERATION=3~"$(lsb_release -cs)"` will produce iterations such as
+`3~trusty`, `3~xenial`, etc. depending on the distro used to build the
+package.
+
+See https://www.debian.org/doc/debian-policy/#version for the official
+Debian package-versioning spec, and
+https://askubuntu.com/questions/620533/how-does-ubuntu-name-packages
+for an explanation of how Ubuntu specifies the `debian_revision`.
```
leandro-lucarella-sociomantic: OK, I like how now it doesn't feel like the only option is the ubuntu scheme, but I still feel this is too verbose and out of place.

First, as a user, if I don't care about this new feature, I don't want to have to scroll 3 pages to have a glimpse of what the next new feature is. This got even longer as before, so more distracting.

Then, is too detailed, to much history. A changelog should be short and to the point. Longer and more descriptive documentation should go in the documentation proper, the readme or whatever. I thought of moving it to the `PKGITERATION` or `iteration` docs in the README. But the README right now is working more like a reference, look at the other variable definitions, they are super short and to the point.

For me it would be super weird that for `PKGVERSION` you'll get:
```rest
* ``PKGVERSION`` is the version to be used when creating packages. It's
   obtained via the ``VERSION`` variable by default.
```
and for `PKGITERATION` a 6 paragraphs-long description with many options of different schemes used by different projects.

For me this kind of description belongs to a tutorial, not to a reference or a changelog, it just dissonant with the rest of the project's docs and changelogs. This project assumes knowledge about packaging, and when you need detailed information, it just points you to another resource with a link to the main source. For me in this case the main source is the Debian docs.

For me this changelog and the README should just say:
```rest
Allows setting the package *iteration* (forwarded to ``fpm`` as ``--iteration``,
known as ``debian_revision``
`in Debian <https://www.debian.org/doc/debian-policy/#version>`_).
```
(in the changelog it would be nice to also mention that before it was hardcoded to `$(lsb_release -cs)`).

I'm sorry, I know you spend some time looking for these sources and it took time and effort to write, and I know is useful information, but again, I think this is not the right place to transmit this information, and for the changelog is counter-productive to be so verbose and for the README it just doesn't go with the style of the rest of the README.

joseph-wakeling-sociomantic: No worries, it's a good point about keeping things simple and brief.  I'll cut things down for this PR and we can follow up with the detailed info in a wiki page or somewhere else.

joseph-wakeling-sociomantic:
Updated with slimmed-down release notes and a tweaked README:

```diff
diff --git a/README.rst b/README.rst
index f7b1c75..6515997 100644
--- a/README.rst
+++ b/README.rst
@@ -415,10 +415,9 @@ Variables you might want to override
   obtained via the ``mkversion.sh`` by default.
 * ``PKGVERSION`` is the version to be used when creating packages. It's
   obtained via the ``VERSION`` variable by default.
-* ``PKGITERATION`` is the iteration of the package to be created (for deb
-  packages, this corresponds to the ``debian_revision``).  By default it is
-  set to the distro code (e.g. ``trusty``, ``xenial``).  It will be ignored
-  if ``PKG_DEFAULTS`` is set manually.
+* ``PKGITERATION`` optionally allows setting the package *iteration* (forwarded
+  to `fpm` as `--iteration`, known as the `debian_revision`
+  [in Debian](https://www.debian.org/doc/debian-policy/#version)).
 * ``PRE_BUILD_D`` and ``POST_BUILD_D`` hold scripts executed before and after
   running the command to build D targets (when using the ``build_d`` function).
   By default they are used to generate the ``Version.d`` file, but users can
diff --git a/relnotes/pkg-iteration.feature.md b/relnotes/pkg-iteration.feature.md
index beda1f3..a84c84d 100644
--- a/relnotes/pkg-iteration.feature.md
+++ b/relnotes/pkg-iteration.feature.md
@@ -6,29 +6,13 @@ deb packages are concerned).
 
 Previously the package iteration was hardcoded to match the distro code
 (e.g. `xenial`), disallowing typical patterns referencing the packager
-and changes to the packaging rather than the upstream project.  For
-example:
-
-* Debian packages use a simple number for the `debian_revision` (1, 2,
-  3, ...), resetting to 1 when the upstream version changes
-
-* Ubuntu packages typically use an `XubuntuY` format where `X` is the
-  numerical revision of the Debian package that has been forked (or `0`
-  if there is no such package), and `Y` is the Ubuntu package revision
-  (hence Ubuntu package versions like `2.7.0-2ubuntu3.1`, where `2.7.0`
-  is the upstream version and `2ubuntu3.1` means Ubuntu revision 3.1 of
-  Debian package revision 2)
-
-* some packagers like to include both revision information and the build
-  or target distro: e.g. packages from the git PPA have revisions such
-  as `0ppa1~ubuntu16.04.1` (note also the leading 0 indicating that this
-  is a new package rather than a fork of a Debian package).
+and changes to the packaging rather than the upstream project.
 
 The new `PKGITERATION` variable allows the package iteration to be set
 to whatever is wanted, whether via `Config.mak` or the command line.
-It defaults to the distro code, meaning nothing will change for existing
-package definitions, but can now be customized where this is appropriate
-(e.g. for packages of 3rd-party projects).
+It still defaults to the distro code, meaning nothing will change for
+existing package definitions, but can now be customized where this is
+appropriate (e.g. for packages of 3rd-party projects).
 
 Note that if you want to include the distro code in a customized package
 iteration, this will have to be included manually.  For example,
@@ -36,7 +20,6 @@ iteration, this will have to be included manually.  For example,
 `3~trusty`, `3~xenial`, etc. depending on the distro used to build the
 package.
 
-See https://www.debian.org/doc/debian-policy/#version for the official
-Debian package-versioning spec, and
-https://askubuntu.com/questions/620533/how-does-ubuntu-name-packages
-for an explanation of how Ubuntu specifies the `debian_revision`.
+For some examples of how to set the package iteration, see the
+ [Debian versioning policy](https://www.debian.org/doc/debian-policy/#version)
+and [the Ubuntu `debian_revision` schema](https://askubuntu.com/questions/620533/how-does-ubuntu-name-packages).
```

joseph-wakeling-sociomantic:
Would you also like tweaks to the commit message?

leandro-lucarella-sociomantic (approved):
Thanks!

leandro-lucarella-sociomantic:
> Would you also like tweaks to the commit message?

Ideally. I don't mind keeping it almost as it is as long as it's said that the ubuntu way is just one possible scheme.

joseph-wakeling-sociomantic:
OK, give me a couple of minutes.

joseph-wakeling-sociomantic:
Commit message updated.  Looks better?

PR #122, opened 2018-02-13
Adapt release notes to new format

leandro-lucarella-sociomantic:


leandro-lucarella-sociomantic:
Self merging as it's trivial and I want to release.

PR #125, opened 2018-06-19
git-deblog: Fix multiple tags for one commit warning

leandro-lucarella-sociomantic:
The warning needed a hash variable that was yet undefined.

### DIFF
--- a/git-deblog
+++ b/git-deblog
@@ -168,14 +168,14 @@ def get_tag(refs):
             tags.remove(t)
     if not tags:
         return None
-    if len(tags) > 1:
-        warn('more than one tag found for commit {}, using the first '
-                'one ({})', hash, ', '.join(tags))
     def get_field(field):
         hdr = field + ' '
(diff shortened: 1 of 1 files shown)


leandro-lucarella-sociomantic:
With neptune pre-releases, it is not uncommon to have 2 tags on the same commit. Also, the first tags is picked, but the last one should be used instead (the final release for example). I will add a fix for that too.

leandro-lucarella-sociomantic:
OK, done, I actually found more potential problems with `git-deblog` (I think pre-releases are not taken into account at all, and sorting would be lexicographical, not semver-based). In any case, I think nobody ever looked at the output of this tool (generating the Debian changelog), so I think probably is not worth any further effort to improve it.

mathias-baumann-sociomantic (approved):

mihails-strasuns-sociomantic (approved):
(not tested)

PR #124, opened 2018-06-19
Save test binaries

leandro-lucarella-sociomantic:
When running tests, since the final target is a `test.stamp` instead of producing the binary (because we care about running them, not just building them), the test binaries themselves are considered intermediate files by Make, and normally removed automatically (even if a test failed).

To avoid this, we need to mark them as `.PRECIOUS:`, for example for integration tests:
```make
.PRECIOUS: $(patsubst $C/integrationtest/%/main.d,$O/test-%,$(wildcard $C/integrationtest/*/main.d))
```

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -580,6 +580,9 @@ $O/%unittests.stamp: $O/%unittests
 		 -p $(call file2module,$(INTEGRATIONTEST)). $(UTFLAGS),$<,run)
 	$Vtouch $@
 
+# Save test binaries in case the user want to re-run them
+.PRECIOUS: $O/%unittests $O/%unittests.d
+
 # Integration tests rules
 ##########################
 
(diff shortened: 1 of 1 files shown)


leandro-lucarella-sociomantic:
Attached fix.

leandro-lucarella-sociomantic:
I guess you didn't actually test it, right? I was about to do a simple test.

leandro-lucarella-sociomantic:
Usually when I assign a PR to myself is because it's ready to be reviewed (and I appreciate approvals) but I want to do the merging myself (because I'm checking something for example).

leandro-lucarella-sociomantic:
My simple test worked, all seems :heavy_check_mark:

mathias-baumann-sociomantic:
I didn't test, no. Will only "approve" next time then :)

PR #129, opened 2019-09-11
Add build option to request position-independent code

joseph-wakeling-frequenz:
Newer distro releases (e.g. Ubuntu 18.04) require builds to use position independent code (i.e. the `-fPIC` build flag), otherwise linking fails.  The new `USE_PIC` build variable can be used to enable this option:

    USE_PIC=1 make [...]

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -183,6 +183,14 @@ override DFLAGS += -O -inline
 endif
 
 
+# Position-independent code
+############################
+
+ifeq ($(USE_PIC),1)
+override DFLAGS += -fPIC
+endif
+
+
 # Directories
 ##############
 
--- a/relnotes/PIC.feature.md
+++ b/relnotes/PIC.feature.md
@@ -1,0 +1,7 @@
+### Position-independent code can now be specified with `USE_PIC`
+
+Newer distro releases (e.g. Ubuntu 18.04) require builds to use position
+independent code (i.e. the `-fPIC` build flag), otherwise linking fails.
+The new `USE_PIC` build variable can be used to enable this option:
+
+    USE_PIC=1 make [...]


Geod24:
If it's required by your distro, wouldn't `/etc/dmd.conf` (or similar) be a better place ?

matthias-wende-sociomantic:
> If it's required by your distro, wouldn't /etc/dmd.conf (or similar) be a better place ?

I agree. Having it as a environment variable seems not to be a clean approach, given that you can always add it to the `Build.mak`. I think that it can also be quite confusing to users if there are to many places  to set options. (e.g. in this case environment variable, compiler config file, Build.mak, DFLAGS env variable).

joseph-wakeling-frequenz:
There is a problem if one wants to use `dmd-transitional`, since this is built only on 16.04 and doesn't have that in its `dmd.conf` by default.  Thus far patches to d**node apps' `Build.mak` have been accepted, so I thought it might be an idea to DRY that up and make it simpler for anyone who wanted to migrate to 18.04 machines.

But if you'd rather keep it app by app in `Build.mak`, fine by me.

Geod24:
> since this is built only on 16.04 and doesn't have that in its dmd.conf by default.

So it's a packaging problem. Is there no 18.04 specific package, or is it missing the switch in the `dmd-transitional.conf` ? DH is still on 16.04, correct?

In any case, this is not a build system problem: whatever build tool you use (MakD, dub, rdmd), you will need to explicitly pass the fPIC switch if you do it at the build system level.

joseph-wakeling-frequenz:
AFAIK `dmd-transitional` is built only on 16.04, hence the lack of the flag.  I believe but obviously cannot confirm that 16.04 is still used for deployment ;-)

The `USE_PIC` var was always intended as a bit of a dirty transitional workaround, but I'd assumed it was an acceptable one since patches to similar effect were merged for individual apps.

Anyway, happy to drop if there is a consensus that this is the wrong way to go.

matthias-wende-sociomantic:
> I believe but obviously cannot confirm that 16.04 is still used for deployment ;-)

I can confirm. But thanks to your help there are chances it's getting updated soon.
I suggest lets do a workaround it on app level until DH updates the linux distro.

joseph-wakeling-frequenz:
OK, good with me.  Then can I get an eye on https://github.com/sociomantic-tsunami/dmqnode/pull/112 please? :-)

Closing here.

PR #130, opened 2020-01-06
Allow building with F=prod

andrew-stevenson-sociomantic:
Requiring the abbreviation of `development` but not allowing the abbreviation
of `production` has been cacthing me for years. This change allows you to use
either form.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -28,6 +28,8 @@ PROJECT_NAME ?= $(shell basename $T)
 
 # Define the valid flavors
 VALID_FLAVORS := devel production
+# Aliases for the above (in the same order)
+VALID_FLAVOR_SHORT := development prod
 
 # Flavor (variant), can be defined by the user in Config.mak
 F ?= devel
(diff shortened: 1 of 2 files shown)


Makd.mak
```diff
@@ -28,6 +28,8 @@ PROJECT_NAME ?= $(shell basename $T)
 
 # Define the valid flavors
 VALID_FLAVORS := devel production
+# Aliases for the above (in the same order)
+VALID_FLAVOR_SHORT := development prod
```
Geod24, line 32: Shouldn't those be inverted ? Setting `VALID_FLAVORS` to `development production` and the short to `devel prod` ?

ibuclaw: Yeah, that makes more sense to me too.

Geod24: I'll fix the PR up soon-ish

llucax: It would be a breaking change, but I would just support the short terms, what's the use for the long versions?

Geod24:
Replaced by #132

PR #132, opened 2020-08-25
[BREAKING] Rename production FLAVOR to prod

Geod24:
Consistent with the short name given to development.
See PR #130 for a previous attempt.

### DIFF
--- a/Makd.mak
+++ b/Makd.mak
@@ -27,7 +27,7 @@ R := $(subst $T,,$(patsubst $T/%,%,$(CURDIR)))
 PROJECT_NAME ?= $(shell basename $T)
 
 # Define the valid flavors
-VALID_FLAVORS := devel production
+VALID_FLAVORS := devel prod
 
 # Flavor (variant), can be defined by the user in Config.mak
 F ?= devel
@@ -59,6 +59,12 @@ EXAMPLE ?= example
 # Check flavours
 FLAVOR_IS_VALID_ := $(if $(filter $F,$(VALID_FLAVORS)),1,0)
 ifeq ($(FLAVOR_IS_VALID_),0)
+
+# The next 3 lines should be removed in v4.x.x
+ifeq ($F,production)
+$(error F=$F is not a valid flavor, 'production' was renamed to 'prod', please update your code.)
+endif
+
 $(error F=$F is not a valid flavor (options are: $(VALID_FLAVORS)))
 endif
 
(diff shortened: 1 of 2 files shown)


Makd.mak
```diff
@@ -27,7 +27,7 @@ R := $(subst $T,,$(patsubst $T/%,%,$(CURDIR)))
 PROJECT_NAME ?= $(shell basename $T)
 
 # Define the valid flavors
-VALID_FLAVORS := devel production
+VALID_FLAVORS := devel prod
```
llucax, line 30: Maybe be nice to the user and check if the value is `production` and fail telling the user to please update it to `prod`?

llucax: Also it should probably target v3.x.x if it is breaking...

Geod24: Ah, created the branch but didn't target it 🙈

Geod24: > Maybe be nice to the user and check if the value is production and fail telling the user to please update it to prod?

Is there any MakD user outside of the projects we maintain ?

llucax: Not that I know of, but even if it's highly unlikely, once you head this road of "nobody is using it" is hard to make any sense to having proper versioning at all IMHO...

llucax: For one side, you'll get an error message saying the flavor is wrong, and if you are a good library user you would have read the changelog (missing!) so it should be obvious what happened, but I still think is only 3 extra lines of code and it makes users experiences nicer if you add something like:
```make
# This should be removed in v4.x.x
ifeq ($F,production)
$(error F=$F is not a valid flavor, 'production' was renamed to 'prod', please update your code.
endif
```
after https://github.com/sociomantic-tsunami/makd/blob/v2.x.x/Makd.mak#L59 (untested).

This can be even added as a deprecation (change `error` to `warning`) to v2.x.x and support both as a migration path.

Geod24: Wouldn't those 3 lines break someone defining `production` as a valid flavor though (or wanting to) ?

llucax:
Yes, good point, you could check if production is in the valid flavours first but maybe is getting too overkill. 🤷‍♂️

llucax:
Anyway, just for the records, I'm not opposing to the change, I just prefer another approach, more for philosophical reasons than practical ones, so I'm fine if you merge it.

Makd.mak
Geod24: > untested

Check your parenthesis!

Geod24:
Updated with the message, moved inside the `if` so someone can override it, should be better now.
Also briefly tested with Ocean.

llucax (approved):

PR #134, opened 2020-09-22
Upgrade beaver, remove v1.x.x mentions, and leftover from D1 days

Geod24:
Hopefully the last bit before releasing v3.0.0.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -219,7 +219,7 @@ Once you have the basic setup done, you can already enjoy a lot of small cool
 features. For example you get a nice, terse and colorful output, for example::
 
         mkversion src/Version.d
-        rdmd1 build/devel/bin/someapp
+        rdmd build/devel/bin/someapp
 
 If there are any errors, messages will appear in red so they are easier to spot.
 
@@ -1176,6 +1176,4 @@ course). There is a convenience target to just clean coverage reports:
 
 Merging and overrideable directory versions:
 
-* ``dmd1``: tangort v1.8.0+
-* ``dmd-transitional``: v2.070.2.s15+ and v2.071.2.s4+
 * ``dmd``: v2.078.0+
--- a/README.rst
+++ b/README.rst
@@ -31,10 +31,8 @@ Maintained Major Branches
 ====== ==================== ===============
 Major  Initial release date Supported until
 ====== ==================== ===============
-v1.x.x v1.4.0_: 24/06/2016  Fri 08/12/2017
-v2.x.x v2.0.0_: 08/09/2017  TBD
+v2.x.x v2.0.0_: 08/09/2017  31/12/2020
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -31,10 +31,8 @@ Maintained Major Branches
 ====== ==================== ===============
 Major  Initial release date Supported until
 ====== ==================== ===============
-v1.x.x v1.4.0_: 24/06/2016  Fri 08/12/2017
 v2.x.x v2.0.0_: 08/09/2017  TBD
 ====== ==================== ===============
-.. _v1.4.0: https://github.com/sociomantic-tsunami/makd/releases/tag/v1.4.0
```
llucax, line 37: Wouldn't it be a good opportunity to set the end of life of V2 too? Given the circumstances I guess a regular period of months is not really necessary.

llucax (approved):
Other than this minor thing, that can be even done in a separate PR, LGTM.

README.rst
Geod24: So, immediately?

llucax: I mean, it is almost unmaintained anyway. Also, the fact that it is unsupported doesn't really mean that you can backport some fix if it is easy and useful enough, but *guaranteeing* a ~6~ 3 months maintenance period (or so) seems excessive given the current state of the project, IMHO. I would make only a *best-effort* promise. :shrug:

llucax: Also, it in unlikely that in the next 3 months a severe bug is found that it is really important to get backported, so in practice for bug fixes it doesn't seem like a big deal. New features or backward compatible changes have a slight higher chance, IMHO, and in this case if v2 is still maintained, new features should go there, which means then need to be merged to v3 and make merge trains, etc. That also looks like an unnecessary overhead.

Also, it probably makes sense to reduce the maintenance guarantee for minor version to maintaining only the latest one.

llucax:
Should this be finally merged?

Version.d.tpl
```diff
@@ -113,16 +101,8 @@ private void buildPredefinedVersions()
...
-        mixin("static immutable Id = `version (` ~ name ~ `) version_info[\"ver_`
-             ~ name ~ `\"] = \"` ~ name ~ `\";`;");
-    }
-    else
-    {
-        const _tpl_istring Id = `version (` ~ name ~ `) version_info["ver_` ~ name ~
-            `"] = "` ~ name ~ `";`;
-    }
+    static immutable Id = `version (` ~ name ~ `) version_info[\"ver_`
+         ~ name ~ `\"] = \"` ~ name ~ `\";`;
```
joseph-wakeling-frequenz, line 107: I think the `\` should have been removed from this string -- they were needed because of the surrounding `mixin("...")` but now they insert actual backslashes.

PR #136, opened 2020-11-30
Remove invalid backslashes from Version.Id template string

joseph-wakeling-frequenz:
The original code for this `static immutable` declaration was wrapped in a `mixin("...")`, so the backslashes were necessary to escape the double quote marks.  However, when D1 support (and hence the mixin) was dropped the backslashes were left behind, and now form a part of the `Id` string (meaning the resulting code is not valid D).

Removing these backslashes restores the `Id` string to its correct form.

### DIFF
--- a/Version.d.tpl
+++ b/Version.d.tpl
@@ -103,6 +103,6 @@ private void buildPredefinedVersions()
 
 private template Id (string name)
 {
-    static immutable Id = `version (` ~ name ~ `) version_info[\"ver_`
-         ~ name ~ `\"] = \"` ~ name ~ `\";`;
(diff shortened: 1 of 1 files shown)


llucax (approved):
LGTM

llucax:
(I leave it to you if you want to merge or wait for @Geod24 )

Geod24 (approved):

joseph-wakeling-frequenz:
@Geod24 want to tag & release a v3.0.1 or shall I ... ?

Geod24:
I'd be happy if you could do it, and I'll update my Ocean PR after that

